PR #62, opened 2021-02-09
Migrate to SDK V2

ClaireBellivier:


oboukili:
@ClaireBellivier do you need any help debugging the failed tests? Have you got any leads yet?

ClaireBellivier:
> @ClaireBellivier do you need any help debugging the failed tests? Have you got any leads yet?

Thanks, for now, I'm adding some temporary logs to figure out what's the first failure is about.

oboukili:
I think it's related to the wait parameter. And more precisely about this change in the SDK v2 https://www.terraform.io/docs/extend/guides/v2-upgrade-guide.html#clearer-handling-of-nil-for-helper-resource-nonretryableerror-and-helper-resource-retryableerror
I guess this call is the culprit [`return resource.NonRetryableError(resourceArgoCDApplicationRead(d, meta))`](https://github.com/oboukili/terraform-provider-argocd/blob/784cd0b12e91d193c50141ce0a072acc9d4ed8c3/argocd/resource_argocd_application.go#L117) . The resourceArgoCDApplicationRead returns a nil error (and should be), but as I read it above it seems it's not expected. I wonder what can we do about that.

Good luck and thanks!

oboukili:
I would probably look into a provider that has successfully migrated to the sdk v2 such as the google provider, and investigate how they deal with that usecase.

oboukili:
I believe I have fixed the issue, my original assumption was correct, the Terraform official docs are therefore seemimgly [incorrect](https://www.terraform.io/docs/extend/resources/retries-and-customizable-timeouts.html#retry) since the general availability of the SDK v2.

Could you handle removing all deprecated function calls and replace them with their Context aware equivalents as stated in the upgrade [documentation](https://www.terraform.io/docs/extend/guides/v2-upgrade-guide.html#more-support-for-context-context)? Thanks!

oboukili (changes requested):
What's left would to replace the deprecated function calls to complete this migration, since all tests are now passing.

ClaireBellivier:
> What's left would to replace the deprecated function calls to complete this migration, since all tests are now passing.

Thanks, it's done ✅

argocd/resource_argocd_application.go
```diff
@@ -20,7 +21,7 @@ func resourceArgoCDApplication() *schema.Resource {
 		Delete: resourceArgoCDApplicationDelete,
```
oboukili, line 21: CRUD must be replaced by their context aware equivalents here

argocd/resource_argocd_project.go
```diff
@@ -19,7 +20,7 @@ func resourceArgoCDProject() *schema.Resource {
 		Delete: resourceArgoCDProjectDelete,
```
oboukili, line 20: CRUD must be replaced by their context aware equivalents here

argocd/resource_argocd_application.go
```diff
@@ -92,11 +93,15 @@ func resourceArgoCDApplicationCreate(d *schema.ResourceData, meta interface{}) e
 	}
 	d.SetId(app.Name)
 	if wait, ok := d.GetOk("wait"); ok && wait.(bool) {
-		return resource.Retry(d.Timeout(schema.TimeoutCreate), func() *resource.RetryError {
+		err = resource.RetryContext(context.Background(), d.Timeout(schema.TimeoutCreate), func() *resource.RetryError {
 			a, err := c.Get(context.Background(), &applicationClient.ApplicationQuery{
 				Name: &app.Name,
 			})
 			if err != nil {
+				fmt.Printf("%v\n", err)
```
oboukili, line 101: You should pass the error on to Terraform error handling, or the new diag.Diagnostics introduced in the sdk v2. Not print an error on the standard output.

argocd/resource_argocd_application.go
```diff
@@ -92,11 +93,15 @@ func resourceArgoCDApplicationCreate(d *schema.ResourceData, meta interface{}) e
...
+				if strings.Contains(err.Error(), "NotFound") {
```
oboukili, line 102: Can you detail that change of behaviour?

oboukili (changes requested):
All CRUD methods should be replaced by their context aware equivalents. I put a few notes on a couple resources to specify which.

argocd/resource_argocd_application.go
ClaireBellivier: > All CRUD methods should be replaced by their context aware equivalents. I put a few notes on a couple resources to specify which.

It is gonna take some time to make those changes, it's probably better to open another PR just for those. Wdyt?

argocd/resource_argocd_application.go
ClaireBellivier: I think we should retry waiting as long as the resource is not found. If another type of error occurs, the wait-loop should stop there.

argocd/resource_argocd_application.go
oboukili: I would prefer to tackle any leftover from the v1 SDK in that PR, and use the v2 SDK updated methods everywhere, so that no deprecated functions are still being called. Then as a second step, adding new features that the v2 SDK offer in another PR - extended TypeMap validator functions come to mind - would be a possibility.

argocd/resource_argocd_application.go
oboukili: That use case should not happen as it is already handled in line 91 to 93, therefore I don't think it's necessary to add that check.

oboukili (changes requested):

ClaireBellivier:
@oboukili the migration to the context functions has been pushed 👍

ClaireBellivier:
It's rebased now and your `cluster` resource has been migrated too. We could have avoided this extra migration if we had split the PR in two.

ClaireBellivier:
Hey, @oboukili a token & timeout issues have been encountered in the CI. After checking, I'm not sure what the root cause is, can you have a look, please?

oboukili:
I reran the test jobs, there are still errors, probably related to how the [new acceptance testing driver](https://www.terraform.io/docs/extend/guides/v2-upgrade-guide.html#a-new-acceptance-testing-driver) is initialized, you should start searching there.

argocd/resource_argocd_project.go
```diff
@@ -60,10 +68,14 @@ func resourceArgoCDProjectCreate(d *schema.ResourceData, meta interface{}) error
 	tokenMutexProjectMap[projectName].RUnlock()
 
 	if err != nil {
-		switch strings.Contains(err.Error(), "NotFound") {
-		case true:
-		default:
-			return err
+		if strings.Contains(err.Error(), "NotFound") {
```
raphink, line 71: This inverts the previous logic, resulting in a failure to create projects. It should be:
```suggestion
		if !strings.Contains(err.Error(), "NotFound") {
```

ClaireBellivier: Done ✅

argocd/resource_argocd_repository.go
```diff
@@ -40,62 +42,92 @@ func resourceArgoCDRepositoryCreate(d *schema.ResourceData, meta interface{}) er
...
-		r, err = c.Get(context.Background(), &repository.RepoQuery{
+		r, err = c.Get(ctx, &repository.RepoQuery{
 			Repo:         d.Id(),
 			ForceRefresh: true,
 		})
 		tokenMutexConfiguration.RUnlock()
 
 		if err != nil {
-			switch strings.Contains(err.Error(), "NotFound") {
-			// Repository has already been deleted in an out-of-band fashion
```
oboukili, line 80: Please keep this comment for justification

argocd/resource_argocd_repository.go
```diff
@@ -40,62 +42,92 @@ func resourceArgoCDRepositoryCreate(d *schema.ResourceData, meta interface{}) er
...
-			case true:
+			if strings.Contains(err.Error(), "NotFound") {
```
oboukili, line 102: Same thing for style consistency, keep the switch block within if block

argocd/schema_cluster.go
```diff
@@ -161,7 +161,6 @@ func clusterSchema() map[string]*schema.Schema {
 					"connection_state": {
 						Type:     schema.TypeList,
 						Computed: true,
-						MaxItems: 1,
```
oboukili, line 164: Why this change?

argocd/structure_application.go
```diff
@@ -284,9 +285,16 @@ func expandApplicationSyncPolicy(_sp []interface{}) (*application.SyncPolicy, er
 			r := _retry[0]
 			for k, v := range r.(map[string]interface{}) {
 				if k == "limit" {
+					var err error
```
oboukili, line 288: Is this change related to error shadowing? Please detail.

go.mod
```diff
@@ -9,7 +9,8 @@ require (
 	github.com/argoproj/pkg v0.2.0
 	github.com/cristalhq/jwt/v3 v3.0.8
 	github.com/golang/protobuf v1.4.3
-	github.com/hashicorp/terraform-plugin-sdk v1.16.0
+	github.com/hashicorp/terraform-plugin-sdk v1.16.1
```
oboukili, line 12: Seems there are still references to sdk v1 if go.mod still contains it

oboukili (changes requested):

argocd/resource_argocd_repository.go
ClaireBellivier: Using `switch` with booleans is quite unusual.

oboukili: Right..

ClaireBellivier:
@oboukili could you please take a look at the last commits which make the acceptance tests fail with this message 
```
resource_argocd_repository_test.go:13: Step 3/5 error: Error running apply: exit status 1
```
There seems to be an incompatibility between the tests setup and the underlying Terraform. Thanks 🙏

oboukili:
Can you answer the previous review comments? Thanks

oboukili:
> @oboukili could you please take a look at the last commits which make the acceptance tests fail with this message
> 
> ```
> resource_argocd_repository_test.go:13: Step 3/5 error: Error running apply: exit status 1
> ```
> 
> There seems to be an incompatibility between the tests setup and the underlying Terraform. Thanks 🙏

Sorry, I don't understand your sentence: `incompatibility between the tests setup and the underlying Terraform`.
The error message seems pretty clear though:
```
         Error: application n18qlu313q could not be created
        
          with argocd_application.public,
          on terraform_plugin_test.tf line 6, in resource "argocd_application" "public":
           6: resource "argocd_application" "public" {
        
        rpc error: code = Internal desc = grpc: error while marshaling: proto:
        required field application.ApplicationCreateRequest.application not set
```

Your library updates (not related to the terraform sdk v2) introduced a non retrocompatible change with argocd 1.x which is used for the tests at the moment.

argocd/resource_argocd_project_token.go
```diff
@@ -100,73 +108,145 @@ func resourceArgoCDProjectTokenCreate(d *schema.ResourceData, meta interface{})
...
+				Summary:  fmt.Sprintf("token claims issue date for project %s is missing", projectName),
+			},
+		}
 	}
 	if expiresInOk {
-		switch claims.ExpiresAt {
-		case nil:
-			return fmt.Errorf("returned expires_at is nil")
-		default:
+		if claims.ExpiresAt == nil {
```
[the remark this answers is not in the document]
ClaireBellivier: cc comment [below here](https://github.com/oboukili/terraform-provider-argocd/pull/62#discussion_r647447267)

argocd/resource_argocd_repository.go
ClaireBellivier: 👍

argocd/schema_cluster.go
ClaireBellivier: It doesn't seem to build with it

```
    provider_test.go:22: err: 1 error occurred:                                                                                                                                                
                * resource argocd_cluster: connection_state: MaxItems is for configurable attributes,there's nothing to configure on computed-only field 
```

argocd/structure_application.go
ClaireBellivier: It doesn't seem to build without it either. The `error` var isn't defined earlier, and we can't use `:=` on the next line.

ClaireBellivier:
> > @oboukili could you please take a look at the last commits which make the acceptance tests fail with this message
> > ```
> > resource_argocd_repository_test.go:13: Step 3/5 error: Error running apply: exit status 1
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > There seems to be an incompatibility between the tests setup and the underlying Terraform. Thanks 🙏
> 
> Sorry, I don't understand your sentence: `incompatibility between the tests setup and the underlying Terraform`.
> The error message seems pretty clear though:
> 
> ```
>          Error: application n18qlu313q could not be created
>         
>           with argocd_application.public,
>           on terraform_plugin_test.tf line 6, in resource "argocd_application" "public":
>            6: resource "argocd_application" "public" {
>         
>         rpc error: code = Internal desc = grpc: error while marshaling: proto:
>         required field application.ApplicationCreateRequest.application not set
> ```
> 
> Your library updates (not related to the terraform sdk v2) introduced a non retrocompatible change with argocd 1.x which is used for the tests at the moment.

What library update do you have in mind? The changes on the `go.mod` are related to the SDK V2 update only. The other update done is running `go mod tidy` to get rid of the SDK V1 obsolete references.

ClaireBellivier:
Left comments are fixed & solved, please check the CI issue for the last one 🙏

oboukili:
This has nothing to do with protobuf, but instead definitely the argocd version. This message below:
```
 required field application.ApplicationCreateRequest.application not set
```

most likely means that the ApplicationCreateRequest function signature has changed and requires a new `application` field.

PR #75, opened 2021-05-10
Create ArgoCD clients on demand

raphink:
This PR aims to create ArgoCD clients on demand, in order to allow setting the
provider dynamically, based on other resource attributes.

The idea is to store the provider's `ResourceData` in the `meta` and use it to
initialize the clients when resources need them.

This PR is based on top of #62.

TODO:

- [x] pass the provider's `ResourceData` as pointer to resources
- [x]  call `initClients()` in all resource types
- [x]  cache clients to avoid duplication


raphink:
@oboukili what is the status for #62?

oboukili:
Unfortunately, there is no answer from the PR author of #62

ClaireBellivier:
If you can wait for next Tuesday, I can have a look at the remaining tasks. Otherwise, feel free to complete the pull request yourself.

raphink:
Great. I'm also happy to help if necessary 😁

raphink:
Stale status: waiting on #62

oboukili:
Hi @raphink , #62 has just been merged, therefore if you could update your branch from the master branch, we can resume the work on this PR.
I will hold off releasing any new version until we merge your PR as this one in addition to #62 warrants a major version bump of the provider.

raphink:
@oboukili Thanks. I'm going to try and rebase. I'll let you know when it's ready.

mcanevet:
@oboukili I tested this PR. It works fine.

mcanevet (approved):

raphink:
@oboukili this should be ready

GNUmakefile
```diff
@@ -1,26 +1,69 @@
...
+	go test $(TEST) -timeout=30s -parallel=4
+
+testacc: fmtcheck
+	TF_ACC=1 go test $(TEST) -v -parallel 20 $(TESTARGS) -timeout 120m
+
+fmt:
+	@echo "==> Fixing source code with gofmt..."
+	gofmt -s -w ./$(PKG_NAME)
+
+# Currently required by tf-deploy compile
```
oboukili, line 46: the script gofmtcheck.sh does not seem to exist, can you add it?

GNUmakefile
```diff
@@ -1,26 +1,69 @@
...
+		echo "  make test-compile TEST=./$(PKG_NAME)"; \
+		exit 1; \
+	fi
+	go test -c $(TEST) $(TESTARGS)
 
-testacc:
-	sh scripts/testacc.sh
+vendor:
+	go mod tidy
+	go mod vendor
```
oboukili, line 65: These helper scripts make directives (testacc*) are useful to spin up a test cluster, at least I use them a lot :) can you leave them within the Makefile?

oboukili (changes requested):

raphink:
@oboukili is it ok like this?

oboukili:
Yes, we just need to wait for the CI tests to pass, I think I noticed a test case error earlier. Let's see.

oboukili:
Apparently there's a race condition 
```
/home/runner/go/pkg/mod/github.com/hashicorp/terraform-plugin-sdk/v2@v2.7.1/helper/schema/resource_data.go:84 +0x5f fp=0xc00395b378 sp=0xc00395b2c8 pc=0x2dad55f
github.com/oboukili/terraform-provider-argocd/argocd.initApiClient(0xc000cecf80, 0x0, 0x0, 0x0, 0x0)
	/home/runner/work/terraform-provider-argocd/terraform-provider-argocd/argocd/provider.go:148 +0x2c6 fp=0xc00395b520 sp=0xc00395b378 pc=0x32081c6
github.com/oboukili/terraform-provider-argocd/argocd.(*ServerInterface).initClients(0xc000b599f0, 0x0, 0x0)
	/home/runner/work/terraform-provider-argocd/terraform-provider-argocd/argocd/features.go:50 +0x8a5 fp=0xc00395b608 sp=0xc00395b520 pc=0x3206de5
github.com/oboukili/terraform-provider-argocd/argocd.resourceArgoCDProjectTokenDelete(0x3d6f130, 0xc001c6e240, 0xc000ceda00, 0x3549560, 0xc000b599f0, 0xc003032e50, 0xc00395b7b0, 0x10fdd78)
	/home/runner/work/terraform-provider-argocd/terraform-provider-argocd/argocd/resource_argocd_project_token.go:420 +0x5b fp=0xc00395b750 sp=0xc00395b608 pc=0x321567b
github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema.(*Resource).delete(0xc001bc3180, 0x3d6f0c0, 0xc0025fdf80, 0xc000ceda00, 0x3549560, 0xc000b599f0, 0x0, 0x0, 0x0)
```

I never had the opportunity to debug this so I can't really offer serious guidance, but I remember `go test -race` might come in handy.

raphink:
I'm guessing this might have to do with multiple resources trying to init the clients at the same time. I could probably add a flag on the clients to mark them as initialized and avoid this behavior.

oboukili:
We should use some kind of locking mechanism such as a sync.RWMutex or try to use a sync.syncMap.
The former will be a bit tedious to add, the latter can be dangerous (never used it) as apparently some guarantees that Go provides get thrown out the window like type safety.

If you use the RWMutex, I think it should be located out of the clients' objects.

raphink:
NOTE: @oboukili I used a `sync.Mutex` instead of a `sync.RWMutex` because we don't want to take the risk that a read-only client would exit the `initClients()` function before the first (write) client has initialized everything.

oboukili:
@raphink the tests timeout quickly now, I believe it's because your lock is an exclusive one, not a RWMutex one, therefore even for a read a lock is positioned. Please try and use a RWMutex lock, thanks.

argocd/features.go
```diff
@@ -43,7 +43,7 @@ type ServerInterface struct {
 	ServerVersionMessage *version.VersionMessage
 	ProviderData         *schema.ResourceData
 
-	InitLock    sync.Mutex
+	InitLock    sync.RWMutex
```
oboukili, line 46: This is not enough, the locking mechanism has to be updated as well in the function with RLock/Lock and RUnlock/Unlock

oboukili (changes requested):

raphink:
@oboukili I'm probably misunderstanding something I guess. If I use an RLock, then some resources will be able to read the `ServerInterface` before it is done initializing. I don't see a way to ensure all resources have fully initialized clients without an exclusive lock. How would I do that?

oboukili:
Extremely naively (sorry I don't have so much time to spare at the moment) I would say:

Modify the initClients() function to:
```go
func (p *ServerInterface) initClients() error {
	if p.Initialized {
		return nil
	}
	p.InitLock.Lock()

	d := p.ProviderData
...
```

and position a RLock very early on on every resource CRUD function:
```go
func resourceArgoCDApplicationCreate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
	objectMeta, spec, diags := expandApplication(d)
	if diags != nil {
		return diags
	}
	server := meta.(*ServerInterface)
	server.InitLock.RLock()
	defer server.InitLock.RUnlock()
       ...
```

Good luck and thanks!

oboukili:
A syncMap would be most interesting though! If applicable.

raphink:
The problem is not really with a map though (or is it?) but with the various fields of the `ServerInterface` struct. Are these managed as a map internally?

raphink:
@oboukili I managed to get it to pass (at least locally) with a simple mutex by just using `defer`, which I hadn't done in the first try. Could you just launch the pipeline to see if it's green like this please?

raphink:
I see the acceptance tests are now passing. @oboukili is that good for you?

oboukili:
Oh great! Awesome work @raphink ! Let me merge that and make a major release ;)

PR #78, opened 2021-05-11
Be explicit about repository dependencies

### LINKED ISSUE #76: Failture to create an argocd_application resource that (implicitly) depends on an argocd_repository
This provider fails to create an argocd_application resource with implicit dependencies on one or more argocd_repository resources. The user should explicitly signal these dependencies to terraform via `depends_on`, but this requirement isn't documented in this provider. I only found [this test example](https://github.com/oboukili/terraform-provider-argocd/blob/v1.2.1/argocd/resource_argocd_repository_credentials_test.go#L65). 

This ticket signals the lack of user documentation on this `depends_on` requirement. For newer users to terraform, it might not be clear that such a dependency signal is needed.

A toy example is included below:
```terraform
resource "argocd_repository" "my-gitops-repo" {
  name            = "hf-gitops"
  repo            = "git@github.com:my-org/my-gitops-repo.git"
  username        = "git"
  ssh_private_key = file("~/.ssh/id_ecdsa")
}

resource "argocd_application" "app-of-apps" {
  metadata {
    name      = "app-of-apps-app"
    namespace = "argocd"
  }

  spec {
    project = "default"

    source {
      repo_url        = "git@github.com:my-org/my-gitops-repo.git"
      path            = "argocd-bootstrap-app"
      target_revision = "master"
    }

    destination {
      server    = "https://kubernetes.default.svc"
      namespace = "argocd"
    }

  depends_on = [argocd_repository.my-gitops-repo]
}

```

Without the `depends_on` statement, `terraform apply` fails as follows:
```
│ Error: rpc error: code = InvalidArgument desc = application spec is invalid: InvalidSpecError: repository not accessible: error creating SSH agent: "SSH agent requested but SSH_AUTH_SOCK not-specified"                                                                                                                    
│                                                                                                                                                                                                                                                                                                                              
│   on main.tf line 26, in resource "argocd_application" "app-of-apps":                                                                                                                                                                                                                                                      
│   26: resource "argocd_application" "app-of-apps" { 
```

ClaireBellivier:
fixes #76

### DIFF
--- a/docs/resources/application.md
+++ b/docs/resources/application.md
@@ -2,9 +2,15 @@
 
 Creates an ArgoCD application.
 
-## Example Usage
+## Example Usages
 
 ```hcl
+resource "argocd_repository" "kustomize-repo" {
+  name            = "kustomize"
+  repo            = "https://github.com/kubernetes-sigs/kustomize"
+  # ...
+}
+
 resource "argocd_application" "kustomize" {
   metadata {
     name      = "kustomize-app"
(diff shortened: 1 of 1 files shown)


oboukili:
No need for that, as explained in #76.

PR #79, opened 2021-05-18
Add support for GRPCWebRootPath ArgoCD ApiClient ClientOption

fvdnabee:
Cfr. argocd docs on running argocd at a different UI root path: [ArgoCD Server and UI Root Path (v1.5.3)](https://argo-cd.readthedocs.io/en/stable/operator-manual/ingress/#argocd-server-and-ui-root-path-v153)


oboukili (approved):

oboukili:
thanks!

PR #108, opened 2021-10-15
support jq path expressions in ignore_difference block

jarrod-manwaring:
This PR adds support for using jq path expressions in an `argocd_application` resources `ignore_difference` block. This [feature](https://github.com/argoproj/argo-cd/pull/6109) was added to ArgoCD in [v2.1.0](https://github.com/argoproj/argo-cd/releases/tag/v2.1.0).

### DIFF
--- a/argocd/features.go
+++ b/argocd/features.go
@@ -20,6 +20,7 @@ import (
 
 const (
 	featureApplicationLevelSyncOptions = iota
+	featureIgnoreDiffJQPathExpressions
 	featureRepositoryGet
 	featureTokenIDs
 )
@@ -27,6 +28,7 @@ const (
 var (
 	featureVersionConstraintsMap = map[int]*semver.Version{
 		featureApplicationLevelSyncOptions: semver.MustParse("1.5.0"),
+		featureIgnoreDiffJQPathExpressions: semver.MustParse("2.1.0"),
 		featureRepositoryGet:               semver.MustParse("1.6.0"),
 		featureTokenIDs:                    semver.MustParse("1.5.3"),
 	}
--- a/argocd/resource_argocd_application.go
+++ b/argocd/resource_argocd_application.go
@@ -104,6 +104,36 @@ func resourceArgoCDApplicationCreate(ctx context.Context, d *schema.ResourceData
 		}
 	}
 
+	featureIgnoreDiffJQPathExpressionsSupported, err := server.isFeatureSupported(featureIgnoreDiffJQPathExpressions)
+	if err != nil {
+		return []diag.Diagnostic{
+			{
+				Severity: diag.Error,
+				Summary:  "feature not supported",
+				Detail:   err.Error(),
+			},
+		}
+	}
+	hasJQPathExpressions := false
+	if spec.IgnoreDifferences != nil {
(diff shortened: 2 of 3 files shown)


jarrod-manwaring:
@oboukili Hello, first time contributor here!

I was hoping you could help me understand why the `v2.1.2` acceptance test might be failing? I expected the other two tests to fail, as the feature this PR adds was not added to ArgoCD until `v2.1.0`, and my local testing confirms as much. However, when I run the acceptance tests locally using `v2.1.2` everything passes.

oboukili:
Hi @jarrod-manwaring , thanks for your contribution! :)

This is really strange if the same test suite works on your end locally (we are talking about the test suite using the terraform test SDK yes? Not just ad-hoc testing, in which case it's perfectly normal you won't see any issues).
This issue usually happens when there is something wrong upon flattening the API response and persisting it to the Terraform state, you should carefully debug what you are persisting and that it matches the type contstraints.

Good luck,

jarrod-manwaring:
@oboukili I think I may have found what was causing the issue. In my local environment I had overwritten the ArgoCD version referenced in the kustomization file [here](https://github.com/oboukili/terraform-provider-argocd/blob/c6ac4a2b457c94566334fe3e21b37fe754683979/manifests/install/kustomization.yml#L7) to `v2.1.2`, which is why my acceptance tests were passing.

Throughout debugging I found that, although the acceptance tests ran via Github Actions seemingly iterate over the [matrix](https://github.com/oboukili/terraform-provider-argocd/blob/c6ac4a2b457c94566334fe3e21b37fe754683979/.github/workflows/tests.yml#L17) of ArgoCD versions to test, each iteration was likely running the acceptance tests using ArgoCD `v1.8.3`, as this version is hardcoded in the kustomization file. Also, in the Github Actions testing configuration, it looks like the proper ArgoCD manifests for the given version are being obtained via this [curl call](https://github.com/oboukili/terraform-provider-argocd/blob/c6ac4a2b457c94566334fe3e21b37fe754683979/.github/workflows/tests.yml#L46) and stored in `manifests/install/install.yml`. This is all well and good, except for the fact that the kustomization file doesn't have `install.yml` added to it's `resources` list, while still having the hardcoded list item for the ArgoCD `v1.8.3` manifests.

The TL;DR is that it looks like CI has been running acceptance tests using only ArgoCD `v1.8.3` for some time now, and builds likely haven't failed as no breaking changes have been committed to the project.

oboukili:
Ouch good find …

jarrod-manwaring:
@oboukili I've added the feature version constraint for jq path expressions, and also broken it out into it's own test step. From there I added a `SkipFunc` to the test step to allow for skipping the test step conditionally, but the way I am determining the ArgoCD server version seems a bit crude. I wanted to avoid relying on a CI environment variable and instead get the version from `ServerInterface`, but this required me to essentially instantiate and configure new `Provider` to get the version. Any ideas on how that could be done more efficiently?

oboukili:
I honestly don’t think it can be done any better than what you did… Unless you want to work on it a bit more, I am going to merge this. I will have to tackle the 2.1 regression with the repository resource before releasing though.

PR #128, opened 2022-01-26
fix(tests): use ssh:// in TestAccArgoCDRepositoryCredentials

raphink:
Arguably, this seems to be a regression in Argo CD itself, so it should probably be forwarded there, but at least this fixes the tests for the provider.

Signed-off-by: Raphaël Pinson <raphael.pinson@camptocamp.com>


oboukili:
well done!

PR #143, opened 2022-02-24
add: project scope support for clusters

alxbse:
Adds support for project scoped clusters: https://argo-cd.readthedocs.io/en/stable/user-guide/projects/#project-scoped-repositories-and-clusters

I'm deliberately adding an acceptance tests that breaks with 2.1.0 because that version does not have support for scoped clusters, any ideas on how to deal with tests that require at least some version of argocd? I'm thinking if adding a precheck function to check the argocd version and skip if it is not compatible could be a good solution?

### DIFF
--- a/docs/resources/cluster.md
+++ b/docs/resources/cluster.md
@@ -121,6 +121,7 @@ resource "argocd_cluster" "eks" {
 * `namespaces` - (Optional) Holds list of namespaces which are accessible in that cluster. Cluster level resources would be ignored if namespace list is not empty..
 * `config` - (Optional) The configuration specification, nested attributes are documented below.
 * `metadata` - (Optional) Cluster metadata, nested attributes are documented below.
+* `project` - (Optional) Scope cluster to ArgoCD project. If omitted, cluster will be global. Requires ArgoCD 2.2.0 onwards.
 
 The `config` block can have the following attributes:
 
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -49,4 +49,6 @@ jobs:
           netstat -tulpn
 
       - name: Run acceptance tests
+        env:
+          ARGOCD_VERSION: ${{ matrix.argocd_version }}
         run: sh scripts/testacc.sh
--- a/argocd/features.go
+++ b/argocd/features.go
@@ -23,6 +23,7 @@ const (
 	featureIgnoreDiffJQPathExpressions
 	featureRepositoryGet
 	featureTokenIDs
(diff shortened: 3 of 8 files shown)


docs/resources/cluster.md
```diff
@@ -121,6 +121,7 @@ resource "argocd_cluster" "eks" {
 * `namespaces` - (Optional) Holds list of namespaces which are accessible in that cluster. Cluster level resources would be ignored if namespace list is not empty..
 * `config` - (Optional) The configuration specification, nested attributes are documented below.
 * `metadata` - (Optional) Cluster metadata, nested attributes are documented below.
+* `project` - (Optional) Scope cluster to ArgoCD project. If omitted, cluster will be global.
```
oboukili, line 124: Maybe add a version requirement comment here, and what happens if the field is added on argocd versions < 2.2

oboukili (changes requested):
Good question about the precheck, easy way would be to ignore silently the project attribute inside the expansion function if the version is below the constraint (and log a debug comment). 
There is a map in features.go that may help you do that.

docs/resources/cluster.md
alxbse: 2.2.0 being required has been added. Also, based on how the application resource does it I added a new feature to the constraints map so if the setting is used with an older version it errors out.

alxbse:
I was looking around in other terraform providers if there are any establish ways of doing prechecks that check server versions but have not been able to find any yet.

I added a slightly hackish precheck function that uses the constraints map to compare what version is required for the acceptance test to run. It uses the ARGOCD_VERSION environment variable to figure out what version it is running against, which is not the nicest solution.

Would it be possible to access the actual provider and the initialized ServerInterface from any function? It would be nicer to get the actual version from the server using the isFeatureSupported function.

alxbse:
Looking at how github.com/hashicorp/terraform-provider-vsphere does it I added an alternative implementation of the precheck function in alxbse/terraform-provider-argocd@2e5f71dd0c98fae938e8503c7ce5930849733ed4. It is still kinda hacky but atleast it uses the same isFeatureSupported function call.

It feels like I'm get further away from the actual issue of adding support for project scope, should I maybe close this PR for now to sort out prechecking in another one?

oboukili:
As you mentioned, we would need to establish another ServerInterface just for the tests. Your solution is acceptable, and the impact on local testing is minimal. For now I think that will do, we can indeed tackle this issue in another PR.

oboukili (approved):

oboukili (approved):

PR #150, opened 2022-03-09
Add local config support

blakepettersson:
This adds the ability to get authentication settings from a local file. This is useful in cases where a user has previously logged in using SSO. The default path is `$HOME/.config/argocd/config`, but this can be overridden by either specifying `ARGOCD_CONFIG_PATH` as an env var, or manually with `config_path`.

There has been some changes in how the default local config path is managed between `2.2.x` and `2.3`, so this PR also upgrades argocd to `2.3.0`, along with its dependencies.


oboukili (approved):

blakepettersson:
Thanks a lot! :bow: For some reason I do not see `2.2.6` or its artifacts under releases even though the release action seemed to have run successfully...

oboukili:
Hi, thanks for your contribution :)
Indeed, the goreleaser step takes a bit of time, and I usually do something else in the meantime and tend to forget about it! It’s now released, sorry about that.

Issue #104, opened 2021-10-09
Empty orphaned_resources block crashes provider


macdweller:
If an argocd_project is specified with an empty orphaned_resources block, like so:

```
resource "argocd_project" "project" {
  metadata {
    name        = "foo"
    namespace   = "argocd"
  }
  spec {
    description = "bar"
    source_repos = ["*"]
    destination {
      server    = "https://kubernetes.default.svc"
      namespace = "*"
    }
    cluster_resource_whitelist {
      group = "*"
      kind  = "*"
    }
    orphaned_resources {
    }
  }
}
```

The provider will crash.

MrLuje:
Fixed in 3.0.0

PR #153, opened 2022-03-20
fix: constant orphaned changes if not specified & crash

### LINKED ISSUE #105: Orphaned resources in projects cause constant change.
If argocd_project.spec.orphaned_resources is not specified, terraform will happily create the object.  When you then immediately re-run a plan, it will want to make changes to the block:

```
  # argocd_project.<redacted> will be updated in-place
  ~ resource "argocd_project" "<redacted>" {
        id = "<redacted>"


      ~ spec {
            # (3 unchanged attributes hidden)



          - orphaned_resources {
            }
            # (2 unchanged blocks hidden)
        }
        # (1 unchanged block hidden)
    }
```

This is very annoying, as it will always result in changes occurring.  A lifecycle{ignore_changes} is necessary.  Specifying an empty block is not recommended due to issue #104; it will update just fine with the empty block, but recreating the resource will cause a crash.

MrLuje:
Resolves #105 & #104 

To be on par with the API surface of Argo CD, "orphaned resources monitoring" will be only enabled if it is explicitly declared. Monitoring enabled because of the previous default behavior will be detected as a removal on next apply.

This is a **BREAKING CHANGE** with the current version :

<table>
<tr>
	<td>"Orphaned resources" block specified
	<td> Current version
	<td> This PR
<tr>
	<td> Yes
	<td> Enables monitoring
	<td> Enables monitoring
<tr>
	<td> No
	<td> Enables monitoring
	<td> Disables monitoring
</table>

### DIFF
--- a/argocd/resource_argocd_project_test.go
+++ b/argocd/resource_argocd_project_test.go
@@ -64,6 +64,16 @@ func TestAccArgoCDProject(t *testing.T) {
 					// TODO: check all possible attributes
 				),
 			},
+			{
+				Config: testAccArgoCDProjectSimpleWithoutOrph(name),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttrSet(
+						"argocd_project.simple",
+						"metadata.0.uid",
+						// TODO: check all possible attributes
+					),
+				),
+			},
 		},
 	})
 }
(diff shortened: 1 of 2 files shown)


oboukili (approved):
Good catch!

MrLuje:
I took a look at the failing test **TestAccArgoCDProject_tokensCoexistence** which has **ExpectNonEmptyPlan: true**. I suppose the non-empty plan was due to the orphaned_resources bug and not the behavior being tested.

Here is the plan generated on master :
```yaml
/terraform-provider-argocd/argocd/resource_argocd_project_test.go:72: Step 1/1, expected an error with pattern, no match on: After applying this test step, the plan was not empty.
        stdout:


        Terraform used the selected providers to generate the following execution
        plan. Resource actions are indicated with the following symbols:
          ~ update in-place

        Terraform will perform the following actions:

          # argocd_project.coexistence will be updated in-place
          ~ resource "argocd_project" "coexistence" {
                id = "test-acc-zozru04kea"

              ~ metadata {
                  + annotations      = {}
                  + labels           = {}
                    name             = "test-acc-zozru04kea"
                    # (4 unchanged attributes hidden)
                }

              ~ spec {
                  + signature_keys = []
                    # (2 unchanged attributes hidden)


                  - orphaned_resources {
                    }

                  ~ role {
                      + groups   = []
                        name     = "testrole"
                        # (1 unchanged attribute hidden)
                    }
                    # (1 unchanged block hidden)
                }
            }

        Plan: 0 to add, 1 to change, 0 to destroy.
```

If this is ok with you, I'll remove the **ExpectNonEmptyPlan** from this test

oboukili:
Hi, ah yes good catch again! Please go ahead!

PR #164, opened 2022-05-01
docs: add missing bearer_token config

chap-dr:
Following the "[Example Usage - GCP GKE cluster](https://registry.terraform.io/providers/oboukili/argocd/latest/docs/resources/cluster#example-usage---gcp-gke-cluster)" guide cluster creation kept failing, Comparing clusters created through the provider with existing clusters showed that the bearer_token config was missing. 

I'm thinking its a mistake since the kubernetes_secret data source is already in the example but unused. 

If for some reason this is by design, or i missed some "before you make PR's here please read and do this and that section" i apologize and the PR can be closed.

### DIFF
--- a/docs/resources/cluster.md
+++ b/docs/resources/cluster.md
@@ -85,6 +85,7 @@ resource "argocd_cluster" "gke" {
   name   = "gke"
 
   config {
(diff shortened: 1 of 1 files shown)


oboukili:
Whoops! Nice catch thanks 😊

Issue #84, opened 2021-07-01
Error: Provider produced inconsistent result after apply


erviker:
argocd version: 2.0.4
terraform version: 1.0.1
agrocd provider version: 1.2.1
kubernetes version: 1.19.10-gke.1600

agrocd.tf
```
provider "argocd" {
  server_addr = "argocd.example.com:443"
  auth_token  = var.agro_token
  insecure = true
}

resource "argocd_cluster" "cluster-1-stage" {
  server = format("https://%s", module.cluster1-stage.endpoint)

  config {
    tls_client_config {
      ca_data = base64decode(module.cluster1-stage.cluster_ca_certificate)
      cert_data = base64decode(module.cluster1-stage.client_certificate)
      key_data = base64decode(module.cluster1-stage.client_key)
    }
  }
}
```
Error
```
argocd_cluster.cluster-1-stage: Creating...
╷
│ Error: Provider produced inconsistent result after apply
│
│ When applying changes to argocd_cluster.cluster-1-stage, provider "provider[\"registry.terraform.io/oboukili/argocd\"]" produced an unexpected new value: Root resource was present, but now absent.
│
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
```

jabouchleih:
Same issue here: 

> │ Error: Provider produced inconsistent result after apply
> │ 
> │ When applying changes to argocd_repository_credentials.argocd_template_repo_creds, provider "provider[\"registry.terraform.io/oboukili/argocd\"]" produced an unexpected new
> │ value: Root resource was present, but now absent.
> │ 
> │ This is a bug in the provider, which should be reported in the provider's own issue tracker.

Resource is simply defined as: 
```
resource "argocd_repository_credentials" "argocd_template_repo_creds" {
  url      = var.repository_url 
  username = var.repository_user
  password = var.repository_pw
}
```
where `var.repository_url` is a GIT HTTPS Url: https://gitlab.com/someproject/somepath/aws-native-argocd-templates.git

Matroxt:
Same here but with a `argocd_repository` ressource.

```
argocd version: 2.0.4
terraform version: 1.0.2
agrocd provider version: 1.2.1
kubernetes version: 1.18.0
```

wolfganghuse:
same here when adding a Cluster...

jabouchleih:
@wolfganghuse @Matroxt 

[...]

PR #165, opened 2022-05-01
fix: Allow cluster name to be omitted & ensure unique cluster on creation

MrLuje:
* Name was already an optional property (defaulted to `server`) but made a weird `ID` = `server/name` (ex: https://kubernetes.default.svc.cluster.local/https://kubernetes.default.svc.cluster.local) which broke the ID parser on read and made further operations impossible.
Should fixes parts of #84 

* I also noticed while testing clusters that if you declare 2 clusters with the same server adress, they will manage the same item on k8s and overwrite each other properties (since we have `upsert: true`). 
I made a few tests on the REST api to make sure I got the same results :
```json
{
  "config": {
    "bearerToken": "abcdef.0123456789abcdef",
    "tlsClientConfig": { "insecure": true }
  },
  "name": "hello",
  "server": "https://kubernetes.default.svc.cluster.local"
}
```
POST => 200 OK

```diff
{
  "config": {
    "bearerToken": "abcdef.0123456789abcdef",
    "tlsClientConfig": { "insecure": true }
  },
-  "name": "hello",
+ "name": "hello2",
  "server": "https://kubernetes.default.svc.cluster.local"
}
```
POST => 400 
```json
{
    "error": "existing cluster spec is different; use upsert flag to force update; difference in keys \"ID,Name\"",
    "code": 3,
    "message": "existing cluster spec is different; use upsert flag to force update; difference in keys \"ID,Name\""
}
```

I added a check before cluster creation to make sure the server was not already created (and a mutex to avoid concurrent create in the same apply)

Again, a bit of "not sure why in Go", I used `client.List` for this check because I got **permission denied** when using `client.Get` (which should have been better):
```go
existingCluster, err := client.Get(ctx, &clusterClient.ClusterQuery{
		Server: cluster.Server,
})
```

### DIFF
--- a/argocd/provider.go
+++ b/argocd/provider.go
@@ -27,6 +27,9 @@ var apiClientConnOpts apiclient.ClientOptions
 // Used to handle concurrent access to ArgoCD common configuration
 var tokenMutexConfiguration = &sync.RWMutex{}
 
+// Used to handle concurrent access to ArgoCD clusters
+var tokenMutexClusters = &sync.RWMutex{}
+
 // Used to handle concurrent access to each ArgoCD project
 var tokenMutexProjectMap = make(map[string]*sync.RWMutex, 0)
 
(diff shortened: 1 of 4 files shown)


oboukili:
Nice! I’ll try to find some time this week to review all that, been slacking off a bit, apologies!

oboukili (approved):

PR #203, opened 2022-09-07
Update ArgoCD dependency to v2.4.x

sl1pm4t:
Requires removal of ksonnet support, which has been removed from ArgoCD itself.

The motivation for upgrading the API is to get support for managing the new RBAC types introduced in v2.4.0 - i.e.`exec` and `logs`. 

Currently when I try create RBAC rules for these types:

```hcl
resource "argocd_project" "proj" {
...
  spec {
...
    role {
      name = "${var.project_name}-admin"
      policies = [
        "p, proj:${var.project_name}:${var.project_name}-admin, exec, create, ${var.project_name}/*, allow",
      ]
    }
```

 I get the following error

```
invalid policy rule 'p, proj:warm-escargot:warm-escargot-admin, exec, create, warm-escargot/*, allow': project resource must be: 'applications', not 'exec'
```


oboukili:
Hi there, thanks for your contribution. It would seem there are dependencies issues, could you try and iron them out? The tests should run automatically without waiting for a manual approval to run them now.

oboukili:
Superceded by #207

PR #212, opened 2022-10-26
Add compatiblity for "source_namespaces" introduced in ArgoCD v2.5.0

karlschriek:
ArgoCD 2.5.0 has just been released.

One of the features included therein, is the ability to deploy ArgoCD Applications to namespaces outside of the controller's own namespace (which is usually `argocd`). In order to prevent permission escalation, a new field "source_namespaces" was added to the AppProject resource, which constrains namespaces within which Applications for that AppProject can be deployed and still be picked up by the conroller.

A full discussion on the feature can be found here: https://github.com/argoproj/argo-cd/pull/9755

This PR updates the schemas and other relevant specs by adding support for "source_namespaces".

It also bumps up the argocd-cd Go package to:

`github.com/argoproj/argo-cd/v2 v2.5.0`

In order to accomodate this, I had to also manually bump the pinned k8s.io dependencies, i.e. `k8s.io/api => k8s.io/api v0.23.1` becomes `k8s.io/api => k8s.io/api v0.24.2`  and so forth.

An issue remains though, as there seems to be some kind of incompatibility with `github.com/argoproj/gitops-engine@v0.7.3`, which is still the latest version.

`go build` will cause:

```
# github.com/argoproj/gitops-engine/pkg/utils/kube
../../go/pkg/mod/github.com/argoproj/gitops-engine@v0.7.3/pkg/utils/kube/resource_ops.go:270:38: not enough arguments in call to k.fact.Validator
        have (bool)
        want (string, *"k8s.io/cli-runtime/pkg/resource".QueryParamVerifier)
../../go/pkg/mod/github.com/argoproj/gitops-engine@v0.7.3/pkg/utils/kube/resource_ops.go:278:30: undefined: resource.NewDryRunVerifier
../../go/pkg/mod/github.com/argoproj/gitops-engine@v0.7.3/pkg/utils/kube/resource_ops.go:326:30: undefined: resource.NewDryRunVerifier
../../go/pkg/mod/github.com/argoproj/gitops-engine@v0.7.3/pkg/utils/kube/resource_ops.go:375:30: undefined: resource.NewDryRunVerifier
```
I do not have a good enough understanding of the underlying dependencies to figure out where exactly this problem originates and how to fix or work around it. I am therefore putting this PR up so long in the hope that someone with more context would be able to quickly identify how to deal with this.


karlschriek:
So I just came across this issue: https://github.com/argoproj/gitops-engine/issues/447, and proceeded to pin the gitops-engine version as follows:

`github.com/argoproj/gitops-engine => github.com/argoproj/gitops-engine v0.7.1-0.20220804190909-2bc3fef13e07`

With this it compiles correctly. Removing WIP tag.

karlschriek:
Ok, I've now added `source_namespaces` entries to the acceptance tests. Updated `scripts/testacc_prepare_env.sh` to set up a test environment that uses ArgoCD v2.5.0. Also made some small updates in the readme so that the images and tags line up exactly with what is in `scripts/testacc_prepare_env.sh`.

Tests all run through for me locally.

karlschriek:
I've run into an issue in live testing which isn't picked up by the acceptance tests. To read an application it needs to now be referenced as `[namespace]/[application]`, unless the application exists in the root namespace (e.g. `argocd`), in which case it can be referenced just as `[application]`. 

In other words, queries such as this:

```
	app, err := c.Get(ctx, &applicationClient.ApplicationQuery{
		Name: &objectMeta.Name
	})
```

will have to be updated to:

```
	app, err := c.Get(ctx, &applicationClient.ApplicationQuery{
		Name: &objectMeta.Name,
		AppNamespace: &objectMeta.Namespace,
	})
```
I didn't build any acceptance tests that specifically deploy to other namespaces, so the defaults ones still worked fine. I'll see if I can find all these queries and also add some more acceptance tests. I'll need to alter the dev environment as well though, since server and application-controller need to be deployed with specific flags in order to enable this feature.

karlschriek:
I accidentally added source_namespaces to schemas V0 and V1. Have removed it now.

karlschriek:
I have now bumped the Go version to 1.19. This resolves the issue with not being able to find the the right go package. Acceptance tests now actually run, but there are some failures. I don't really get why the "argocd_cluster" tests are failing, unless there is something fundamentally different in 2.5.0 there that I haven't accounted for...

karlschriek:
Have now added back the tests on the older ArgoCD versions. I am pretty sure I am missing something here. How do I set it up so that certain tests only run for certain versions?

And am I actually supposed to create a new applicationSpecSchema (v3), since the spec has now has an additional field?

onematchfox:
Just seen that I missed a bunch of comments and additional work since I starting writing that comment - which now isn't necessarily valid. Hopefully my reference to `featureProjectScopedClusters` points you in the right direction. Otherwise, rest assured this is on my radar and I'll try take a closer look again as soon as I have the chance.

karlschriek:
@onematchfox yeah, the "argocd_cluster" cluster test problem seems to have gone away. There are other things failing though, in particular on the older versions, although 2.5.0 also fails (although in the latest run it fails on something to do with TLS, for which there is a comment that this doesn't always work).

I think it is almost there now, but two things in particular I am having a hard time figuring out and it might actually be more productive if someone closer to the project takes a look at them:
1. If and how to create a new version for the AppProject schema. Right now I've _edited_ the V2 schema by adding "source_namespaces" to it, but I presume what should have happened instead is that a V3 along with an "upgrade" function gets created.
2. Certain tests will only work on v2.5.0+, but I haven't been able to figure out how to not have them used in earlier versions.

I've also just _removed_ the `source_namespaces` entries again from all the AppProject tests. I had initially added it to all of them, but have now instead created just one dedicated entry. This revealed another problem, which is that I should have set source_namespaces up to be an optional parameter.

Anyway, I am going to leave it there for now.

karlschriek:
@onematchfox have you had a chance to look at this yet?

onematchfox:
> @onematchfox have you had a chance to look at this yet?

HI @karlschriek,

Apologies for the delay. I have been trying to look at this when/where possible but between travelling and falling ill I'm afraid I haven't had much time to dedicate to it. 

> 1. If and how to create a new version for the AppProject schema. Right now I've edited the V2 schema by adding "source_namespaces" to it, but I presume what should have happened instead is that a V3 along with an "upgrade" function gets created.
> 2. Certain tests will only work on v2.5.0+, but I haven't been able to figure out how to not have them used in earlier versions.
> 
> I've also just removed the source_namespaces entries again from all the AppProject tests. I had initially added it to all of them, but have now instead created just one dedicated entry. This revealed another problem, which is that I should have set source_namespaces up to be an optional parameter.

I've pushed a commit that fixes up some of the changes to the project schema although this is by no means complete yet. To specifically address your points. 
1. There is no need to creation a new schema version since, as you later point out, this needs to be an optional field. 
2. I've added the required code (to the project tests at least) that does the required validation of the ArgoCD version to determine whether or not to source namespaces is supported. 

Outstanding from my side at the moment is wrapping my head around the rest of the changes. Namely:
- updates to the application resource and associated tests - will need similar treatment to the project resource/tests
- the changes you've made to the setup script including: 
  - the addition of `testdata_v2.5.0+/token_resource.yml`) - my gut feel is that it is going to be better to inline the `test` project in the application tests
  - the updates to the kustomize overlays
- addressing potential issues with the changes I made to the project resource to handle defaults for source namespaces when running against older ArgoCD versions

onematchfox:
@karlschriek Think I finally managed to get everything working here. 

@oboukili, do you want to take a look at this to double-check I haven't missed anything? One thing worth noting is that I had to adjust the structure of the internal ID for the `argocd_application` resource to store both the `name` and `namespace` in order to support importing resources that have been created in custom/non-default namespaces.

onematchfox (approved):
@oboukili Do you want to double-check this one or can I go ahead and merge?

argocd/schema_project.go
```diff
@@ -576,6 +576,11 @@ func projectSpecSchemaV2() *schema.Schema {
 					Required: true,
 					Elem:     &schema.Schema{Type: schema.TypeString},
 				},
+				"source_namespaces": {
```
oboukili, line 579: While this works, and I am arriving very very late to the party (sorry!), I wonder if `TypeSet` of Strings would not have been preferable here so that a change in the list elements' ordering will not retrigger a resource update, but it changes a lot of the logic underneath, and would imply some additional work. Therefore, I'm fine with letting it as is.
Look at how `jq_path_expressions` is handled for reference.

argocd/schema_application.go
```diff
@@ -1228,3 +1614,15 @@ func resourceArgoCDApplicationStateUpgradeV1(_ context.Context, rawState map[str
 
 	return nil, fmt.Errorf("error during state migration v1 to v2, 'ksonnet' support has been removed")
 }
+
+func resourceArgoCDApplicationStateUpgradeV2(_ context.Context, rawState map[string]interface{}, _ interface{}) (map[string]interface{}, error) {
+	_metadata, ok := rawState["metadata"].([]interface{})
+	if !ok || len(_metadata) == 0 {
+		return nil, fmt.Errorf("failed to read metdata during state migration v2 to v3")
```
oboukili, line 1621:
```suggestion
		return nil, fmt.Errorf("failed to read metadata during state migration v2 to v3")
```

argocd/schema_application.go
```diff
@@ -1166,6 +1166,383 @@ func applicationSpecSchemaV2() *schema.Schema {
 	}
 }
 
+func applicationSpecSchemaV3() *schema.Schema {
```
oboukili, line 1169: This new schema does not seem to be used anywhere, unless I'm mistaken?

argocd/schema_application.go
```diff
@@ -1184,6 +1561,15 @@ func resourceArgoCDApplicationV1() *schema.Resource {
...
 
+func resourceArgoCDApplicationV2() *schema.Resource {
+	return &schema.Resource{
+		Schema: map[string]*schema.Schema{
+			"metadata": metadataSchema("appprojects.argoproj.io"),
+			"spec":     applicationSpecSchemaV2(),
+		},
+	}
+}
+
```
oboukili, line 1572: It's my understanding that the Application schema does not change, but now the `metadata.namespace` field can be set to a value other than `argocd` therefore, why not keep using `applicationSpecSchemaV1()` ? Unless I'm missing something !

```suggestion
func resourceArgoCDApplicationV2() *schema.Resource {
	return &schema.Resource{
		Schema: map[string]*schema.Schema{
			"metadata": metadataSchema("appprojects.argoproj.io"),
			"spec":     applicationSpecSchemaV1(),
		},
	}
}
```

oboukili (changes requested):

oboukili:
@onematchfox big thanks for handling this PR! @karlschriek we're getting there!

argocd/resource_argocd_application_test.go
```diff
@@ -1659,6 +1694,60 @@ resource "argocd_application" "crds" {
 	`, name)
 }
 
+func testAccArgoCDApplicationCustomNamespace(name string) string {
+	return fmt.Sprintf(`
+resource "argocd_project" "simple" {
+  metadata {
+    name      = "%[1]s"
```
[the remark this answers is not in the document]
onematchfox: :thinking: Not sure I follow you here. There are 3 parameters in the format string. The 1st and 3rd are the project name (and so should be the same) whereas the 2nd is the application name. So, I could change this to something like 

``` go
fmt.Sprintf(`
resource "argocd_project" "simple" {
  metadata {
    name      = "%[1]s"
...
resource "argocd_application" "simple" {
  metadata {
    name      = "%[2]s
    namespace = "mynamespace-1"
  }
  spec {
    project = "%[1]s"
...
`, projectName, applicationName)
```

Whereas, if I were to change to using `%s` then I would need to repeat the `name` parameter it would look something like the following (unless there's some syntax that I'm missing here?)

``` go
fmt.Sprintf(`
resource "argocd_project" "simple" {
  metadata {
    name      = "%s"
...
resource "argocd_application" "simple" {
  metadata {
    name      = "%s
    namespace = "mynamespace-1"
  }
  spec {
    project = "%s"
...
`, name, name, name)
```

Also, the option using `%s` is now strictly incorrect as parameter 1 and 3 could be different. And so, since we don't really care if the project name and the application name are different, I took the easy (lazy?) way out and opted for the more terse form of making them all the same rather than option 1. Happy to adjust to either if you would still prefer that over the current implementation?

argocd/schema_project.go
onematchfox: This is a great point. I took a look and I don't think it will take much to rework this implementation so I'll get on it.

argocd/schema_application.go
karlschriek: Yes, that is my understanding as well. It is only the AppProject schema that changes (with the addition of the optional field, "source_namespaces").

argocd/schema_application.go
onematchfox: Oops. Yes, this isn't used. I should have updated the following as well https://github.com/oboukili/terraform-provider-argocd/blob/24699dfe067dd4116512c1b7bb6f7158ffbb476d/argocd/resource_argocd_application.go#L30

For now I have removed this but it may be worth reviewing this decision in light of the comment/discussion below.

argocd/schema_application.go
onematchfox: You are correct that the schema itself has not changed in this PR. However, we do need to do a [state migration](https://github.com/oboukili/terraform-provider-argocd/blob/24699dfe067dd4116512c1b7bb6f7158ffbb476d/argocd/resource_argocd_application.go#L56-L60) to ensure that the `Id` on existing resources is updated to include the namespace. For this to happen, my understanding is that we need to upgrade the [schema version](https://github.com/oboukili/terraform-provider-argocd/blob/24699dfe067dd4116512c1b7bb6f7158ffbb476d/argocd/resource_argocd_application.go#L44)  on the application resource. The application resource currently makes uses [applicationSpecSchemaV2](https://github.com/oboukili/terraform-provider-argocd/blob/24699dfe067dd4116512c1b7bb6f7158ffbb476d/argocd/resource_argocd_application.go#L30) (which dropped support for `ksonnet` in the spec schema (V1). So, in this case I believe the code I have implemented is correct. 

That said, I do wonder how important alignment of version numbers between the schema for the `spec` field of the resource and the schema of the resource itself is? Is it easier for someone reading the code if the schema of the `spec` field

https://github.com/oboukili/terraform-provider-argocd/blob/24699dfe067dd4116512c1b7bb6f7158ffbb476d/argocd/resource_argocd_application.go#L30

aligns with the schema of the resource itself

https://github.com/oboukili/terraform-provider-argocd/blob/24699dfe067dd4116512c1b7bb6f7158ffbb476d/argocd/resource_argocd_application.go#L44

My initial thought was to ensure that this aligned (hence the initial addition of `applicationSpecSchemaV3`). However, this isn't strictly necessary since there are no changes to the `spec` field. So, now we are in a situation where we are using `V2` of the `spec` schema in `V3` of the resource schema. Are you good with this or would you prefer if everything aligned? (Hopefully I've explained this well enough :crossed_fingers:)

argocd/resource_argocd_application_test.go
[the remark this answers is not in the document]
oboukili: I fully agree, but in this case of the `func testAccArgoCDApplicationCustomNamespace(name string) string
` function (line 1697), only _one_ parameter is provided: `name`. Therefore format indexing is indeed not needed ;)

argocd/schema_application.go
oboukili: I understand. How about making a little simplification then?
```go
func applicationSpecSchemaV2() *schema.Schema {
  # To support project source namespaces, we do need to do a state migration to ensure that the Id on existing resources is updated to include the namespace. 
  # For this to happen, we need to trigger a dummy schema version upgrade on the application resource.
  return applicationSpecSchemaV1()
}
```

oboukili (changes requested):

argocd/resource_argocd_application_test.go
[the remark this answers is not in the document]
onematchfox: > I didn't fully read your comment, my sincerest apologies. It was indeed more satisfactory with the %[1]s notation, feel free to roll this change back. Again my apologies! bow

You did have me very confused :smile: No worries. Have dropped that commit now.

oboukili (approved):

Issue #173, opened 2022-05-18
ArgoCD provider wipes `notified.notifications.argoproj.io` notifications on updates


luke-plausin-kidsloop:
Hi there,

I have a problem with my setup that is causing a lot of duplicate notifications to get sent when using the Terraform ArgoCD provider.

It seems that the ArgoCD notifications project uses an annotation - `notified.notifications.argoproj.io` to track whether it has already sent out notifications for a particular app. The annotation is added directly to the app itself, and does not appear on any Terraform diffs. When an update is made through the Terraform ArgoCD provider, this annotation seems to get wiped out by the provider, and then recreated by the argocd notification service (with a duplicate sent out). This is causing a lot of duplicate notifications to get sent, and defeats the point of having the `notified.notifications.argoproj.io` annotation.

It seems that a merge request was accepted a while back to ignore all `notified.notifications.argoproj.io` annotations within the provider. The reason was to try and stop irrelevant details from always appearing in the terraform diff.

https://github.com/oboukili/terraform-provider-argocd/pull/71

I believe that this is what causes the annotation to get wiped. I think the approach of ignoring the annotation altogether is wrong, and the reporter of that issue could have used a Terraform lifecycle statement if they did not want the ArgoCD notifications annotation to appear in their diff.

```
  lifecycle {
    ignore_changes = [
      metadata[0].annotations["notified.notifications.argoproj.io"]
    ]
  }
```

Is it possible to preserve this annotation in some way? Even if it does appear in the diff I think it's better to be explicit about these changes in Terraform updates rather than to silently ignore and then change the values. I'm also keen to hear if there are any workarounds that could help.

Thanks

PR #224, opened 2022-12-19
Finalizers on Application Resource

jdmaguire:
I'm deploying an ArgoCD Application with this provider.

No one _should_ destroy it outside of using this module. Which is good because my application needs a cascading delete for the cleanup. (Which is the default behaviour with this provider.)

On the off chance when someone [does run `kubectl delete` without patching in the finalizer,](https://argo-cd.readthedocs.io/en/stable/user-guide/app_deletion/#deletion-using-kubectl) I want the app to be cleaned up properly.

To do this, I need to set the finalizer when I create the app. This adds that the ability to specify the finalizer at app creation.


jdmaguire:
I think the tests will fail because of bitnami/redis 16.9.11 has been purged from their git repo.

[They regularly purge the bitnami helm repo.](https://raw.githubusercontent.com/bitnami/charts/archive-full-index/bitnami) and that chart version has been removed.

jdmaguire:
> I think the tests will fail because bitnami/redis 16.9.11 has been purged from their helm repo.
> 
> [They regularly purge the bitnami helm repo.](https://github.com/bitnami/charts/issues/10833) and that chart version has been removed.

Fixed in this PR: https://github.com/oboukili/terraform-provider-argocd/pull/224#issuecomment-1358321477

onematchfox:
Rebased this after the merge of #225 so we can see what the tests look like. Haven't had the chance to look at the code at all though.

jdmaguire:
> Rebased this after the merge of #225 so we can see what the tests look like. Haven't had the chance to look at the code at all though.

Merged in master.

argocd/resource_argocd_application_test.go
```diff
@@ -424,6 +424,53 @@ func TestAccArgoCDApplication_Recurse(t *testing.T) {
...
+						"metadata.0.finalizers.#",
+						"1",
+					),
+					resource.TestCheckResourceAttr(
+						"argocd_application.simple",
+						"metadata.0.finalizers.0",
+						"resources-finalizer.argocd.argoproj.io",
+					),
+				),
+			},
```
onematchfox, line 465: Please add test step to verify importing resources as well. E.g.
https://github.com/oboukili/terraform-provider-argocd/blob/3531a32e53d7f12a5fd7f6a2868bc0ceda83fea1/argocd/resource_argocd_application_test.go#L45-L50

onematchfox (changes requested):
Hi @jdmaguire,

Finally got around to looking at this one in a little more detail. I spotted a few places that will require some updates. Most notably, because the `metadata` schema is shared across a number of resources this change has a slightly larger _blast_ radius that you may have initially expected. 

On a more general note, @oboukili, I'd like a second opinion on the overall scope of the work. I can't escape the nagging feeling that tracking `finalizers` within Terraform may be a bad idea. Generally, one would expect that these would be mutated by server-side components (i.e. outside of Terraform) which, would lead to perpetual diffs in the Terraform plan. I believe this is why, for example, the Kubernetes Provider for Terraform does not track these. I'd also be interested to know if you remember considering this when initially implementing the `metadata` schema? 

FWIW, in the context of the ArgoCD resources it looks like tracking `finalizers` may be OK; reading the docs it would seem like ArgoCD does not automatically add any finalizers:

> When you invoke argocd app delete with --cascade, the finalizer is added automatically.

I.e. `finalizers` would not be mutated as part of the normal lifecycle (prior to deletion). But could there potentially be any other actors that may be interesting in these resources who may have mutated that field? Or, would this be considered an edge case worthy of a user having to explicitly add a `lifecycle { ignore_changes = [ ... ] }` block to avoid a perpetual diff?

Edit: P.S. @oboukili how are docs currently managed in this repo? Are manual updates required or are these generated?

jdmaguire:
Thanks for the comments and tips Brian. It will be a day or two until I have time to incorporate the suggestions.

oboukili:
So, there's an exhaustive proposal that should give a hint about the future SSA support in ArgoCD [1], they even begun rolling out some of these proposals around March of 2022 [2]. 
 
However, all of that does not apply in our case, as these features are for resources managed by an ArgoCD application itself.

I believe we should keep things simple (i.e. don't interact directly with Kubernetes) and safe, in order to only own and track the finalizer on the given Application (or Project, Cluster, etc) by implementing changes in this PR in order not to overwrite/delete any finalizers that we don't handle explicitly in the TF spec.

This is indirectly related to #173, which underlined that metadata data tracking should be handled with extreme care: we shipped an implementation of a less than ideal design there, and I believe we should find a generic solution for this type of issues.

I also foresee that there will be some implementation challenges as our Terraform tests might not be happy: there might be State diffs during the tests in between reads for example, but we have to find a solution to make this work.

[1] https://argo-cd.readthedocs.io/en/stable/proposals/server-side-apply/
[2] https://argo-cd.readthedocs.io/en/stable/user-guide/sync-options/#server-side-apply

onematchfox:
Thanks for all the investigation, @oboukili. I also came across that proposal on SSA and came to the same conclusion as you in that it applies solely to resources managed by ArgoCD. I have not looked at the code, but I also don't expect this to affect ArgoCD resources created via the API. And so, whilst I'm not a fan of mutating finalizers on the client side, this does seem to be the expected behaviour in this case. Further evidence to support this can be seen in the behaviour of application sets - which [explicitly set the `finalizers`](https://argo-cd.readthedocs.io/en/stable/operator-manual/applicationset/Application-Deletion/) on any applications that are created (unless configured not too).

One consideration that I did have here when going through this is whether, instead of adding `finalizers` to the metadata schema (which will affect multiple resources), we should leverage the existing [`cascade`](https://github.com/oboukili/terraform-provider-argocd/blob/94394f54556644ee165f1e0d5df24bfe47e31716/argocd/resource_argocd_application.go#L37-L42)[^cascade] field on `applications` and use that to ensure the expected `finalizer` is set? Or is that mixing concerns and introducing too much magic into the provider?

[^cascade]: incidentally, this field is currently undocumented, but this will be addressed in #229

PR #189, opened 2022-06-03
Feat/adding argocd repository project scoped

felipem1210:

PR #226, opened 2022-12-30
feat: add support for add project scoped repositories

### LINKED ISSUE #183: Support creation of project scoped argocd repositories and clusters
Since argocd v2.2.0 there is the possibility to add clusters and repositories project scoped, using the `--project` flag, when using the cli, for example:

`ARGOCD_AUTH_TOKEN=my-project-token argocd repo add --name stable https://gitlab.com/myrepo.git --type git --project my-project`

I see that using the provider to create the repository there is no option to pass the project.

The proposal is to add something like this in the repository resource:

`resource "argocd_repository" "myrepo" {
  repo = "https://gitlab.com/myrepo.git"
  name = "my-repo"
  type = "git"
  project = "my-prokect
}
`
So if project argument is passed create the repository project scoped based.

This same can be applied for clusters

MrLuje:
Cherry-picked @felipem1210 commits from #189 (thanks for initial work !)
* added feature checks (repository project available since v2.2.0)
* added tests

Resolves #183

### DIFF
--- a/argocd/provider_test.go
+++ b/argocd/provider_test.go
@@ -77,6 +77,6 @@ func testAccPreCheckFeatureNotSupported(t *testing.T, feature int) {
 		t.Fatal("feature constraint is not handled by the provider")
 	}
 	if i := versionConstraint.Compare(serverVersion); i != 1 {
-		t.Skipf("version %s does not support feature", v)
+		t.Skipf("not running test if feature is already supported (v%s)", v)
 	}
 }
--- a/argocd/resource_argocd_repository_test.go
+++ b/argocd/resource_argocd_repository_test.go
@@ -98,7 +98,7 @@ func TestAccArgoCDRepositoryScoped(t *testing.T) {
 					),
 					resource.TestCheckResourceAttr(
 						"argocd_repository.helm",
-						"project_name",
+						"project",
 						projectName,
 					),
 				),
@@ -154,7 +154,7 @@ resource "argocd_repository" "helm" {
 `)
 }
 
-func testAccArgoCDRepositoryHelmProjectScoped(project_name string) string {
+func testAccArgoCDRepositoryHelmProjectScoped(project string) string {
 	return fmt.Sprintf(`
 	resource "argocd_project" "simple" {
 	metadata {
@@ -177,9 +177,9 @@ resource "argocd_repository" "helm" {
   repo = "https://helm.nginx.com/stable"
   name = "nginx-stable-scoped"
   type = "helm"
-  project_name = "%s"
+  project = "%s"
 }
-`, project_name, project_name)
+`, project, project)
 }
 
 func testAccArgoCDRepositoryPublicUsageInApplication(name string) string {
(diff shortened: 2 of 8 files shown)


argocd/resource_argocd_repository_test.go
```diff
@@ -99,6 +154,34 @@ resource "argocd_repository" "helm" {
...
+	}
+
+resource "argocd_repository" "helm" {
+  repo = "https://helm.nginx.com/stable"
+  name = "nginx-stable-scoped"
+  type = "helm"
+  project = "%s"
+}
+`, project, project)
+}
```
onematchfox, line 183: Nit: formatting + use of format indexing to ensure resulting string is always correct.

``` suggestion
resource "argocd_project" "simple" {
  metadata {
    name      = "%[1]s"
    namespace = "argocd"
  }

  spec {
    description  = "simple project"
    source_repos = ["*"]
    
	destination {
      name      = "anothercluster"
      namespace = "bar"
    }
  }
}

resource "argocd_repository" "helm" {
  repo = "https://helm.nginx.com/stable"
  name = "nginx-stable-scoped"
  type = "helm"
  project = "%[1]s"
}
`, project)
}
```

argocd/provider_test.go
```diff
@@ -77,6 +77,6 @@ func testAccPreCheckFeatureNotSupported(t *testing.T, feature int) {
 		t.Fatal("feature constraint is not handled by the provider")
 	}
 	if i := versionConstraint.Compare(serverVersion); i != 1 {
-		t.Skipf("version %s does not support feature", v)
+		t.Skipf("not running test if feature is already supported (v%s)", v)
```
onematchfox, line 80: Not sure I follow the logic here? Tests should be run if the feature is supported. They should be skipped when the feature is not supported.

onematchfox (changes requested):
Couple minor nits but otherwise looks good.

argocd/resource_argocd_repository_test.go
MrLuje: didn't know the indexing syntax, nice :)

argocd/provider_test.go
MrLuje: It is used in this test : https://github.com/oboukili/terraform-provider-argocd/pull/226/files#diff-6e254abd82d715b01ed5073e07acb0324eb3dc17fe7e13d9c0f6134b85cc26e3R110-R137 (making sure the "not supported" error is raised)
Since we are using a matrix during tests, there is no point to run this test if the version supports it (IMHO, the previous wording was misleading)
While writing this, I noticed the v2.2.0 is out of tested versions, so the test wasn't needed at all

onematchfox: > IMHO, the previous wording was misleading

You are quite correct - my bad. Was mixing up this function, `testAccPreCheckFeature**Not**Supported`, with `testAccPreCheckFeatureSupported`.

onematchfox (approved):

PR #228, opened 2023-01-11
Use `schema.TypeSet` for application `sync_policy` 

onematchfox:
This PR changes the type of `application.spec.sync_policy.automated` and  `application.spec.sync_policy.retry.backoff` from `schema.TypeMap` to `schema.TypeSet`. 

**Motivation:**

When using `schema.TypeMap` it is not possible to define the set of keys required as per the [provider SDK docs](https://developer.hashicorp.com/terraform/plugin/sdkv2/schemas/schema-types#typemap):

> Using the Elem block to define specific keys for the map is currently not possible.

This leads to some subtle issues. For example, if users do not define all keys this will result in a perpetual diff since all keys are stored in state when flattening the resource. This is illustrated in the fix to the tests containing in https://github.com/oboukili/terraform-provider-argocd/commit/913b530d89c64277416d6b61e0182084ef00bbd5.

Additionally, the use of maps prevents us from [generating docs](https://developer.hashicorp.com/terraform/registry/providers/docs#generating-documentation) for the provider since the keys are not defined. 

**Impact**:

Obviously this is a breaking change to existing consumers of the provider. Although, as illustrated in the documentation updates the upgrade path is a simple one. 

``` terraform
resource "argocd_application" "this" {
  ...

    sync_policy {
      automated = {
        prune = true
      }
      retry {
        backoff = {
          max_duration = "2m"
        }
      }
    }
}
```

would change to 

``` terraform 
resource "argocd_application" "this" {
   ... 

    sync_policy {
      automated {
        prune = true
      }
      retry {
        backoff {
          max_duration = "2m"
        }
      }
    }
}
```


oboukili:
That's great, using TypeMap at the time was mainly because it allowed for faster iteration (as you can imagine through what you experienced with this PR). Great stuff!
Since it's introducing a breaking change, let's first merge the outstanding PRs.

oboukili (approved):

oboukili:
Hi @onematchfox , since the other PRs are in a standby state, feel free to merge and release this as a new major version, since there is a breaking change (make sure to specify this in the release notes). Thanks!

PR #230, opened 2023-01-19
Support passing mtls certs as values

jmcconnell26:
Hi!

We recently ran into a scenario where we wanted to use this argocd provider to authenticate to an ArgoCD instance behind mTLS from Terraform Cloud.

To achieve this, we had to pass the certificate values as variables, rather than paths to the certificate.

Please let me know if there is anything you would like me to change!

Thanks,
Josh


oboukili:
Hi, thanks for your contributions. The PR title is actually misleading, since this is not related to mTLS, but rather certificate-based authentication (not necessarily _mutual_ TLS). And from what I understood from the code, the PR allows passing a certificate (cert,key, cacert) as string values parameters (only file paths are currently supported).

This can already be done outside of this provider using the `local_file` and `local_sensitive_file` resources [1], and passing their paths to the ArgoCD TF provider.

I am definitely not a fan writing to the filesystem, there are multiple things to take into account, like multi-tenancy security (access control should be set at the filesystem level) and simultaneous invocations (think of a long running CI worker, every tenant should have a dedicated temporary folder to host the written files, which may be different), and OS differences (you are assuming the tenant is running a Unix OS, what about Windows?), and cleanup (you don't want to leave unencrypted keys unattended). This would add a lot complexity.

Therefore, for all these reasons, and because there's currently a workaround (even though it is not ideal), I think your PR should not be merged. Let me know if you want to discuss this further.

[1] https://registry.terraform.io/providers/hashicorp/local/latest/docs/resources/sensitive_file

jmcconnell26:
@oboukili, many thanks for the detailed explanation! 

I had been unaware of the existing workaround, which definitely makes more sense to use for this scenario.

PR #235, opened 2023-03-02
Add support for GitHub app credentials

philamente:
We need to switch our authentication towards github to GitHub App. As there was also a feature request for this one in the issues, I hence added GitHub App functionality to the following ressources:

- argocd_repository
- argocd_respository_credentials

The underlying package already [supports](https://pkg.go.dev/github.com/argoproj/argo-cd/v2/pkg/apis/application/v1alpha1#RepoCreds) it

### DIFF
--- a/argocd/structure_repository.go
+++ b/argocd/structure_repository.go
@@ -93,10 +93,10 @@ func flattenRepository(repository *application.Repository, d *schema.ResourceDat
 		"type":                          repository.Type,
 		"githubapp_enterprise_base_url": repository.GitHubAppEnterpriseBaseURL,
 	}
-	for repository.GithubAppId > 0 {
+	if repository.GithubAppId > 0 {
 		r["githubapp_id"] = convertInt64ToString(repository.GithubAppId)
 	}
-	for repository.GithubAppInstallationId > 0 {
+	if repository.GithubAppInstallationId > 0 {
 		r["githubapp_installation_id"] = convertInt64ToString(repository.GithubAppInstallationId)
 	}
 	for k, v := range r {
--- a/argocd/structure_repository.go
+++ b/argocd/structure_repository.go
@@ -49,6 +49,18 @@ func expandRepository(d *schema.ResourceData) *application.Repository {
 	if v, ok := d.GetOk("type"); ok {
 		repository.Type = v.(string)
 	}
+	if v, ok := d.GetOk("githubApp_id"); ok {
+		repository.GithubAppId = v.(int64)
+	}
+	if v, ok := d.GetOk("githubApp_installation_id"); ok {
+		repository.GithubAppInstallationId = v.(int64)
+	}
+	if v, ok := d.GetOk("githubApp_enterprise_base_url"); ok {
+		repository.GitHubAppEnterpriseBaseURL = v.(string)
+	}
+	if v, ok := d.GetOk("githubApp_private_key"); ok {
+		repository.GithubAppPrivateKey = v.(string)
+	}
 	return repository
 }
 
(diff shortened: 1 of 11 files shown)


onematchfox (changes requested):
Hi @philamente,

Thanks for the contribution! 

There are a few things that need to be addressed before we can merge this.

1. I see you have added these fields to both the structure of `repository` and `repository_credentials` but have only added them to the schema of `repository_credentials` (and not `repository`). Was this intentional? 
2. Please can you ensure that you add tests covering the population of the newly added attributes to the affected resources? Note: at present, the acceptance tests are failing, so this will also need to be addressed, although I think this will be resolved by point 1.
3. As `GithubAppId` and `GithubInstallationId` are `int64` types within ArgoCD, I believe you will need to set these to `schema.TypeString` within the Terraform schema (rather than `schema.TypeInt`) and then convert between string/int64 as appropriate. This can be tested/validated by testing the max boundary of these values in the test you add in point 2. For reference, see the `shard` attribute on the `argocd_cluster` resource and the usage of the `utils` function `convertStringToInt64` within this repository.
4. Docs will need to be updated based to reflect changes in point 3.

Feel free to reach out if anything I've said is unclear and/or if you need more specific guidance!

philamente:
Hi @onematchfox, 

thanks for your kind feedback! I saw the failing tests from github actions - unfortunately my local test setup did not yet work properly (working on a corporate pc with annoying limitations - might switch to private on for testing env). 

This contribution is my first attempt in Go, so thanks for baring with me and providing such a detailed feedback. ;) I also noticed that we might have an issue with one of the maps which were set to String only and hence will break with int. Your proposal in 3) could circumvent that.

Thanks!

philamente:
@onematchfox  Maybe you have an idea: the two ids (github_app as well as the github_app_installation) require type int64. Hence using the Int64Pointer does not work/make sense.

However through the conversion to String the optional (unset) case of 0 does not get handled as null by terraform as it gets parsed to "0".

--> Right now I am thinking about the following:
1) Inserting a modified intToString conversion in .utils that handles that optional 0 case

But most likely I miss something here, so maybe you have an even better/more elegant solution

image

argocd/structure_repository.go
```diff
@@ -64,13 +84,16 @@ func flattenRepository(repository *application.Repository, d *schema.ResourceDat
...
-		//"tls_client_cert_key":     repository.TLSClientCertKey,
-		"tls_client_cert_data": repository.TLSClientCertData,
-		"type":                 repository.Type,
+		//"password":                		repository.Password,
+		//"ssh_private_key":         		repository.SSHPrivateKey,
+		//"tls_client_cert_key":     		repository.TLSClientCertKey,
+		"tls_client_cert_data":          repository.TLSClientCertData,
+		"type":                          repository.Type,
+		"githubapp_id":                  convertInt64ToString(repository.GithubAppId),
+		"githubapp_installation_id":     convertInt64ToString(repository.GithubAppInstallationId),
```
onematchfox, line 95: To answer your question in https://github.com/oboukili/terraform-provider-argocd/pull/235#issuecomment-1456441202... I would conditionally set these keys on the map based on the value of `repository.GithubAppId` and `repository.GithubAppInstallationId`. E.g. 

```go
if repository.GithubAppId > 0 {
	r["githubapp_id"] = convertInt64ToString(repository.GithubAppId)
}

if repository.GithubAppInstallationId != 0 {
	r["githubapp_installation_id"] = convertInt64ToString(repository.GithubAppInstallationId)
}
```

onematchfox: Had a brief look but don't see the issue - it's weird :) Will have to do some playing around and will try to find the time to do so this weekend.

onematchfox: Hi @philamente,

I managed to figure out what was happening here. This behaviour is a result of _shortcomings_ in the ArgoCD API. The `RepoCredsService` does not implement a `Get` method, so the provider [makes use](https://github.com/oboukili/terraform-provider-argocd/blob/124d87be91701f969512d2b855e868929c0245db/argocd/resource_argocd_repository_credentials.go#L79-L81) of the `ListRepositoryCredentials` endpoint to read data about existing repository credentials. This function only [returns](https://github.com/argoproj/argo-cd/blob/7be094f38d06859b594b98eb75c7c70d39b80b1e/server/repocreds/repocreds.go#L58-L61) the `url` and `username` and so the provider is unable to track drift between what is configured in Terraform versus the actual state of the underlying Kubernetes secret that ArgoCD stores the repo credentials in. As such, I have slimmed down the `flattenRepositoryCredentials` function to reflect this. I've also opened https://github.com/argoproj/argo-cd/issues/12823 to see if we can improve the situation in future.

onematchfox: Also, as I went through the code, I addressed any other issues I noticed. I hope you don't mind me doing this; for your first contribution in Go, you did a great job, and I was hesitant about "denying" you further learning opportunities. However, given your last comment, I figured you wouldn't mind my expediting the process. I've kept the changes I made atomic, so feel free to take a look at the commits I added and let me know if anything is unclear or if you'd like me to explain any of them. Other than that, this is good to go now. So, once you let me know you're happy with the changes and, if there aren't any further changes from your side, I'll approve and merge.

philamente: > Also, as I went through the code, I addressed any other issues I noticed. I hope you don't mind me doing this; for your first contribution in Go, you did a great job, and I was hesitant about "denying" you further learning opportunities. However, given your last comment, I figured you wouldn't mind my expediting the process. I've kept the changes I made atomic, so feel free to take a look at the commits I added and let me know if anything is unclear or if you'd like me to explain any of them. Other than that, this is good to go now. So, once you let me know you're happy with the changes and, if there aren't any further changes from your side, I'll approve and merge.

Hey, first of all thanks for digging into it and identifying the issue with the underlying ArgoCD API. Also I am really appreciating your input via single "topic" commits which indeed helped me to understand the missing/to be improved parts. I do not mind that at all!

--> Happy with it and we can go ahead with approval and merging

P.S.: I feel sorry for this akward "for" instead "if" copy pasta error in the flattening part of structure_repository. Shame on me I missed it there ;)

onematchfox (approved):

argocd/structure_repository.go
onematchfox: > P.S.: I feel sorry for this akward "for" instead "if" copy pasta error in the flattening part of structure_repository. Shame on me I missed it there ;)

All good... We all make silly mistakes at times :smile:

onematchfox:
Going to go ahead and merge this now. Thanks, @philamente, for the contribution!

philamente:
> Going to go ahead and merge this now. Thanks, @philamente, for the contribution!

Thank you very much for feedback and review!

PR #240, opened 2023-03-28
Fix support for application resource on latest versions of ArgoCD

### LINKED ISSUE #247: Unable to create applications when targetting ArgoCD v2.6.7
There seems to be a [breaking](https://github.com/argoproj/argo-cd/issues/13000) change in how Applications are handled in v.2.6.7 of ArgoCD, which causes issues with the provider. I've tested this with 4.x.x versions and also the latest version 5.0.1.

When creating an application, I'm always getting the following error:
```
Error: application feedback could not be created
rpc error: code = PermissionDenied desc = permission denied
```

Previously, I used an older version of ArgoCD and it worked fine there.

onematchfox:
The latest versions of ArgoCD `2.[4|5|6].` introduced a breaking change to the return value of the `Get` endpoint on the API  in order to resolve [GHSA-2q5c-qw9c-fmvq](https://github.com/argoproj/argo-cd/security/advisories/GHSA-2q5c-qw9c-fmvq). As such it is no longer possible to determine the non-existence of an application using this endpoint (see https://github.com/argoproj/argo-cd/issues/13000). 

As per the [suggested workaround](https://github.com/argoproj/argo-cd/issues/13000#issuecomment-1483135268), this PR changes this codebase to rather make use of the `List` endpoint to retrieve application details as this endpoint will still return a NotFound error code is no such application exists. 

Notes: 
- I took a quick look at the ArgoCD codebase to try and confirm that all fields on an application are returned via the `List` endpoint, and it seems like they are (or at least I didn't spot any form of _filtering_ on the return). In addition, all tests are passing, so I don't believe this will cause any issues. However, in the case that this does cause a diff for some attributes (which are likely untested at present) then we may need to re-implement a call to `Get` within `resourceArgoCDApplicationRead` after first asserting the existence of the application using `List`. Obviously this is less efficient so I'm hoping we don't have to do this.

Closes #247


onematchfox:
Tests failing due to fix for https://github.com/argoproj/argo-cd/security/advisories/GHSA-2q5c-qw9c-fmvq. See related issue at https://github.com/argoproj/argo-cd/issues/13000

mcanevet:
@oboukili @MrLuje would it be possible to review and merge this ASAP as this provider is not compatible anymore with the latest release of ArgoCD (2.6.7+). Many thanks

oboukili (approved):

PR #245, opened 2023-03-29
Bug fixes to `argocd_project_token` + addition of `renew_after`

onematchfox:
This PR contains several updates to the `argocd_project_token` resource. Feel free to let me know if you'd like these split out into separate PRs. 

Fixes:
- Fix `inconsistent result after apply` error when updating the value of `renew_before` for an existing token
- Fix implementation of `renew_before` as this field currently does nothing. For background, I [noticed this](https://github.com/onematchfox/terraform-provider-argocd/blob/63e87bd6920bbd693b0f5ae6a4ffad7ec2f8094b/argocd/resource_argocd_project_token.go#L383-L384), when fixing linting errors in #241. It would seem like this functionality was never fully implemented and has never worked since neither `expiresIn` not `renewBefore` were ever set/read from the state. I am, however, a little uncertain as to what the intention here really was. The existing description of this attribute is that the token will be refreshed when `expires_in - renew_before < currentDate`. This doesn't make much sense to me since both `expires_in` and `renew_before` are durations, and the difference between them is always going to be less than `currentDate` (assuming this is converted to UNIX epoch). My interpretation of this is that this should have been `expiresAt - currentDate < renew_before`. I.e. when the remaining token lifetime is less than `renew_before`, then the token should be automatically renewed. @oboukili would love to hear your thoughts on my interpretation vs what you originally had in mind when implementing this. 

Feature:
- adds `renew_after`, which can be used to automatically rotate a token after a period of time has passed. This is particularly useful for non-expiring tokens as it allows for the rotation of these on a _regular_ basis. Personally, I find this mechanism more preferrable than using a combination of `expires_in` and `renew_before` as one doesn't run the risk of a token expiring due to Terraform not being run in that period. 

Notes:
- This set of changes has a subtle side effect that is worth noting. At present existing tokens are not deleted upon rotation. I.e. when a token expires, a new one is created, but the old one is not cleaned up. As part of this change, old tokens will be deleted when the new one is created. 
- See comment in https://github.com/oboukili/terraform-provider-argocd/commit/2c3b6e00d703c6f03179601a4a8a68297a5be6d7 for more information regarding the decision to remove the hard check that logging in with a generated token works as well as the removal of the arbitrary 5 minutes difference between `expires_in` and `renew_before`. Happy to discuss these decisions in more detail if needed. 
- PR also contains changes from #244 (will rebase once that PR is merged)


oboukili:
Hi there, you got it right, it was supposed to silently renew these static tokens upon a TF run once the renewal period was reached - which was expected to be the `[expiration date - window duration, expiration date]`. Even my own text description confused me to be honest, so great job tackling this.

The `renew_after` sounds perfect, really. Perfectly fine with the cleaning up as well, the leftovers were definitely not intended.

oboukili (approved):

PR #248, opened 2023-04-04
Bug fix: interface conversion: interface {} is nil

yair-segal:
Issue: When defining an empty block of `plugin`, in this way: `plugin { }` , terraform exits with an error:

```
panic: interface conversion: interface {} is nil, not map[string]interface {}

goroutine 40 [running]:
github.com/oboukili/terraform-provider-argocd/argocd.expandApplicationSourcePlugin({0xc000eb8330?, 0xc000565d10?, 0x2f6abaf?})
        github.com/oboukili/terraform-provider-argocd/argocd/structure_application.go:96 +0x372

```

This fix is to keep from terraform failing for that reason.

### DIFF
--- a/argocd/structure_application.go
+++ b/argocd/structure_application.go
@@ -90,7 +90,7 @@ func expandApplicationSource(_as interface{}) (
 }
 
 func expandApplicationSourcePlugin(in []interface{}) *application.ApplicationSourcePlugin {
-	if len(in) == 0 {
+	if len(in) == 0  || in[0] == nil {
 		return nil
 	}
 	a := in[0].(map[string]interface{})


onematchfox:
Hi @yair-segal,

Thanks for submitting this. Unfortunately, whilst this change will fix the immediate `panic`, I don't think it is complete with regards to enabling the use of an empty `plugin` block as described. I have a strong suspicion that the use of `plugin { }`, even with this change, will result in a perpetual diff when planning/applying. To confirm/deny this hypothesis, we will need to add tests for this use case. I, unfortunately, have limited experience with config management plugins and no experience with sidecar plugins (which I believe is your use case). Would you be able to set up a locally reproducible example of this use case and add tests to cover it? Let me know if you need pointers to the relevant places in the codebase where I think changes would be needed.

onematchfox:
> yes - please give me pointers to the relevant places.

OK. 

1. I would start by adding an explicit test for this configuration. You can use the sample [here](https://github.com/oboukili/terraform-provider-argocd/blob/fac513dc46132f7f090206d0eac3f9b233339a38/argocd/resource_argocd_application_test.go#L819-L843) (and the associated configuration [here](https://github.com/oboukili/terraform-provider-argocd/blob/fac513dc46132f7f090206d0eac3f9b233339a38/argocd/resource_argocd_application_test.go#L1704-L1756)) as a blueprint. Doing so will confirm whether or not a perpetual diff results from this change.
2. I suspect that in order to even apply an empty plugin block, we will need to configure the local dev environment with a sidecar plugin. If so, then hopefully, you have an example we can use. Or, perhaps the [example in the docs](https://argo-cd.readthedocs.io/en/stable/operator-manual/config-management-plugins/#write-the-plugin-configuration-file) would work?  The local/test environment is setup using [testacc_prepare_env.sh](https://github.com/oboukili/terraform-provider-argocd/blob/master/scripts/testacc_prepare_env.sh) and the [manifests in the `install` folder](https://github.com/oboukili/terraform-provider-argocd/tree/master/manifests/install). So, you could add a `ConfigManagementPlugin` manifest to that and then patch the `argocd-repo-server` `Deployment` accordingly.

Hope this helps!

PR #263, opened 2023-04-24
Fix locking on resources during Terraform actions

onematchfox:
~This PR addresses the flaky cluster tests by fixing the locking mechanism used to control the concurrent provisioning of resources via the provider. The existing mechanism is flawed in that locks are only maintained for individual calls to the API, thereby allowing cases where for example:~
~- a delete may occur in the middle of the update of an existing resource~
~- two creates of identical resources may run at the same time leading to a plan diff~

~Examples of both can be seen in the logs [here](https://github.com/oboukili/terraform-provider-argocd/actions/runs/4789891646/jobs/8519015272).~

~The change in this PR ensures that a single lock is maintained for the full Terraform CRUD action (and all API calls required to complete it). Also, since the same locking pattern is used across a number of resources, I have taken the opportunity to fix this across the codebase rather than just the cluster resource.~

### DIFF
--- a/argocd/resource_argocd_project.go
+++ b/argocd/resource_argocd_project.go
@@ -235,24 +235,31 @@ func resourceArgoCDProjectUpdate(ctx context.Context, d *schema.ResourceData, me
 
 	projectName := objectMeta.Name
 
-	if _, ok := tokenMutexProjectMap[projectName]; !ok {
-		tokenMutexProjectMap[projectName] = &sync.RWMutex{}
-	}
-
-	tokenMutexProjectMap[projectName].RLock()
-	p, err := si.ProjectClient.Get(ctx, &projectClient.ProjectQuery{
-		Name: d.Id(),
-	})
-	tokenMutexProjectMap[projectName].RUnlock()
-
+	featureProjectSourceNamespacesSupported, err := si.isFeatureSupported(featureProjectSourceNamespaces)
 	if err != nil {
 		return []diag.Diagnostic{
 			{
 				Severity: diag.Error,
-				Summary:  "failed to get project",
+				Summary:  "feature not supported",
 				Detail:   err.Error(),
 			},
 		}
+	} else if !featureProjectSourceNamespacesSupported {
+		_, sourceNamespacesOk := d.GetOk("spec.0.source_namespaces")
+		if sourceNamespacesOk {
+			return []diag.Diagnostic{
+				{
+					Severity: diag.Error,
+					Summary: fmt.Sprintf(
+						"project source_namespaces is only supported from ArgoCD %s onwards",
+						featureVersionConstraintsMap[featureProjectSourceNamespaces].String()),
+				},
+			}
+		}
+	}
+
+	if _, ok := tokenMutexProjectMap[projectName]; !ok {
+		tokenMutexProjectMap[projectName] = &sync.RWMutex{}
 	}
 
 	projectRequest := &projectClient.ProjectUpdateRequest{
@@ -262,7 +269,22 @@ func resourceArgoCDProjectUpdate(ctx context.Context, d *schema.ResourceData, me
 		},
 	}
 
-	if p != nil {
+	tokenMutexProjectMap[projectName].Lock()
+
+	p, err := si.ProjectClient.Get(ctx, &projectClient.ProjectQuery{
+		Name: d.Id(),
+	})
+	if err != nil {
+		tokenMutexProjectMap[projectName].Unlock()
+
+		return []diag.Diagnostic{
+			{
+				Severity: diag.Error,
+				Summary:  fmt.Sprintf("failed to get existing project %s", projectName),
+				Detail:   err.Error(),
+			},
+		}
+	} else if p != nil {
 		// Kubernetes API requires providing the up-to-date correct ResourceVersion for updates
 		projectRequest.Project.ResourceVersion = p.ResourceVersion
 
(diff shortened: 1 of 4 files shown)


onematchfox:
Tests passing :tada:

argocd/resource_argocd_cluster.go
```diff
@@ -88,17 +88,16 @@ func resourceArgoCDClusterCreate(ctx context.Context, d *schema.ResourceData, me
 		}
 	}
 
-	client := *server.ClusterClient
+	tokenMutexClusters.Lock()
```
oboukili, line 91: Initially, I wanted to set an exclusive lock as late as possible, not to impact performance when a lot of clusters are being managed by ArgoCD. This will effectively slow down concurrent clusters creation. WDYT?

argocd/resource_argocd_cluster.go
```diff
@@ -263,8 +230,9 @@ func resourceArgoCDClusterUpdate(ctx context.Context, d *schema.ResourceData, me
 	}
 
 	tokenMutexClusters.Lock()
-	_, err = client.Update(ctx, &clusterClient.ClusterUpdateRequest{Cluster: cluster})
-	tokenMutexClusters.Unlock()
+	defer tokenMutexClusters.Unlock()
```
oboukili, line 233: Same thing here, the provider will delay all other ops on unrelated clusters even further, as the `readArgoCDCluster` processing time is now included. I believe this will prove to be a problem at scale, WDYT?

argocd/resource_argocd_project.go
```diff
@@ -66,18 +65,14 @@ func resourceArgoCDProjectCreate(ctx context.Context, d *schema.ResourceData, me
 		}
 	}
 
-	c := *server.ProjectClient
 	projectName := objectMeta.Name
 
-	if _, ok := tokenMutexProjectMap[projectName]; !ok {
-		tokenMutexProjectMap[projectName] = &sync.RWMutex{}
-	}
+	obtainProjectLock(projectName, true)
```
oboukili, line 70: good idea, but I would make so that the obtainProjectLock function returns an unlock function to be called through the defer statement rather to mix both global variables and setters.

argocd/resource_argocd_project.go
```diff
@@ -170,54 +163,21 @@ func resourceArgoCDProjectRead(ctx context.Context, d *schema.ResourceData, meta
...
-		return []diag.Diagnostic{
-			{
-				Severity: diag.Error,
-				Summary:  fmt.Sprintf("project %s could not be flattened", d.Id()),
-				Detail:   err.Error(),
-			},
-		}
-	}
+	obtainProjectLock(projectName, false)
+	defer tokenMutexProjectMap[projectName].RUnlock()
```
oboukili, line 169: same remark as above.

argocd/resource_argocd_project.go
```diff
@@ -238,26 +198,14 @@ func resourceArgoCDProjectUpdate(ctx context.Context, d *schema.ResourceData, me
...
-	}
-
-	projectRequest := &projectClient.ProjectUpdateRequest{
-		Project: &application.AppProject{
-			ObjectMeta: objectMeta,
-			Spec:       spec,
-		},
-	}
+	obtainProjectLock(projectName, true)
+	defer tokenMutexProjectMap[projectName].Unlock()
```
oboukili, line 204: ➕

oboukili (commented):
Great stuff, the addition of a new projects mutex definitely helps! However I am concerned about the added processing time regarding exclusive locks. While it definitely is a no brainer to release *read* locks through defer statements (though a cancel func call would be preferred as opposed to using the global map variable), I believe for specific - large scale - users, it could have bad consequences.

argocd/resource_argocd_project.go
onematchfox: I like that. Done.

argocd/resource_argocd_cluster.go
onematchfox: :thinking:

onematchfox: Been giving this some more thought. This definitely deserves a bit more discussion (and thought). So, for the time being, I figured I would break this PR down into smaller chunks (that don't necessarily require as much thought :smile:). I will provide a longer answer with my thoughts on this specific point ASAP.

oboukili: To be more specific, the lock release timing is crucial. I (vaguely) remember the test suite became instantly faster with the introduction of RLocks, so the assumption remained that TF blueprints handling large amounts of non thread-safe resources (tokens, and now clusters) may reach timeouts upon creation.

I realize this is an edge case, and don't have actual numbers to back that up. I'll leave that up to your discretion.

onematchfox:
> Approving, but with the performance warnings cited above.

Thanks. I will hold off for the time being to give the performance considerations some more thought. My main issue was the flaky cluster tests (which were addressed in #264) since I have spent way too much time rerunning the test suite to get everything passing.

onematchfox:
After further thought, I've decided to leave the existing locking pattern in place and have reduced this PR to addressing specific issues identified on create/update of projects and on the deletion of project tokens. That said, I think that the use of locks within this project should be revised in general. I suspect that there are historic reasons for the existence of this pattern that may no longer apply when using the latest versions of ArgoCD. For example, I can understand the existence of `tokenMutexConfiguration`, which was used to control concurrent access to the underlying `argoc-cm` `ConfigMap`. However, a number of resources that make use of this lock are no longer stored in that `ConfigMap` and instead are stored in their own `ConfigMap` (e.g. repository certificates in the `argocd-tls-cert-cm` `ConfigMap`) and/or individual `Secret`s (repositories and repository credentials). On a related note, I also think we could also do with explicitly dropping support for older versions of ArgoCD (at least `1.x`) and cleaning up the related feature flags. Doing so, would enable us to revisit the need for/use of the various locks across all resources. However, with all that said, this requires a lot more time and thought than I have available for the time being. So, yeah, I've decided to keep this simple and go with, "if it ain't broke, don't fix it".

PR #291, opened 2023-05-31
Refactor error handling

onematchfox:
Simplifies error handling across all resources through the introduction of 3 utils methods (see related discussion [here](https://github.com/oboukili/terraform-provider-argocd/pull/278#discussion_r1204744146)):
- `featureNotSupported` - used to simplify error returned if a user makes use of configuration options that are not supported by their ArgoCD version
- `argoCDAPIError` - used to standardise error messages returned by the provider when handling errors returned by the ArgoCD API during CRUD operations
- `errorToDiagnostics` - provides a terser syntax to return an error message (and detail) when handling errors

**Note**: as part of this change, we are explicitly dropping feature "flags" for all unsupported versions of ArgoCD. In most cases, this won't affect consumers since, if they are using an outdated version of ArgoCD, then they would have received an error when trying to make use of newer configuration options. I.e. we can be confident that they are not making use of features that are not supported for their version of ArgoCD. However, there are two cases where, within the provider, we supported both old and new versions, whereas, from this point on, we will only support the _new_ versions. As such, anyone using the following resources and ArgoCD versions will need to ensure they upgrade their ArgoCD instance:
- `argocd_project_token` with ArgoCD version `<1.5.3`
- `argocd_repository` with ArgoCD version `<1.6.0`

Whilst this could be seen as a breaking change, my inclination is that, due to how outdated the ArgoCD versions are, we simply highlight this change with a warning in the release notes.

### DIFF
--- a/argocd/resource_argocd_project_token.go
+++ b/argocd/resource_argocd_project_token.go
@@ -230,13 +230,7 @@ func resourceArgoCDProjectTokenCreate(ctx context.Context, d *schema.ResourceDat
 	tokenMutexProjectMap[projectName].Unlock()
 
 	if err != nil {
-		return []diag.Diagnostic{
-			{
-				Severity: diag.Error,
-				Summary:  fmt.Sprintf("token for project %s could not be created", projectName),
-				Detail:   err.Error(),
-			},
-		}
+		return argoCDAPIError("create", "token for project", projectName, err)
 	}
 
 	token, err := jwt.ParseString(resp.GetToken())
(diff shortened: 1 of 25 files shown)


argocd/resource_argocd_project_token.go
```diff
@@ -535,48 +372,14 @@ func resourceArgoCDProjectTokenDelete(ctx context.Context, d *schema.ResourceDat
...
-			{
-				Severity: diag.Error,
-				Summary:  "explicit token ID support could be not be checked",
-				Detail:   err.Error(),
-			},
-		}
-	}
-
-	if featureTokenIDSupported {
-		opts.Id = d.Id()
```
onematchfox, line 550: #294 addresses the removal of this line.

Issue #155, opened 2022-03-28
Feature request - `argocd_application` data source and readable attributes for `argocd_application` resource


luke-plausin-kidsloop:
Hi there,

I am using the ArgoCD provider to create and manage ArgoCD apps in a Terraform project and it's working really well.

Now I have a need to conditionally update some resources managed in Terraform upstream, depending on whether an ArgoCD app has synced successfully. Based on the documentation it seems like the current `argocd_application` resource does not publish any readable attributes. It would be really helpful if the application sync status and health could be read from the Terraform resource as attributes. Also an `argocd_application` data resource would be really useful for getting this information from other apps which might have been deployed outside of the current project.

Thanks

PR #295, opened 2023-06-05
Add read-only `status` attribute to `argocd_application`

### LINKED ISSUE #234: get application ressources on sync
Hello!

I tried to create an argo application with this provider, but it seems that the argo application resource output is getting only the application ressource itself.

I was expecting to see the ressources deployed with the application, like when I do an argo app sync myapp or argo app get myapp.

Is it possible to add the ressources deployed with the application (since it's part of the argocd sync output), to make it available in my terraform plan at some point?

onematchfox:
Partial resolution of #155 
Closes #234

Note: unfortunately, due to a known [issue](https://github.com/hashicorp/terraform-plugin-docs/issues/66) with generating docs on nested, computed attributes, the resulting documentation contains no description :disappointed: This can be resolved by [migrating from the provider SDKv2 to the plugin framework](https://developer.hashicorp.com/terraform/plugin/framework/migrating) although that's a whole project in itself :grimacing:

### DIFF
--- a/argocd/resource_argocd_application.go
+++ b/argocd/resource_argocd_application.go
@@ -42,6 +42,7 @@ func resourceArgoCDApplication() *schema.Resource {
 				Optional:    true,
 				Default:     true,
 			},
+			"status": applicationStatusSchema(),
 		},
 		SchemaVersion: 4,
 		StateUpgraders: []schema.StateUpgrader{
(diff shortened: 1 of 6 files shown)


oboukili:
Good god, it took months migrating to sdkv2 a couple years ago...

oboukili (approved):

PR #314, opened 2023-07-03
Add support for timezone in sync window

jsmcnair:
Whilst the ArgoCD API now supports setting the timezone in a project's sync window, the provider does not expose this so I have had to set it manually. This change exposes the timezone capability in the provider so that it can be configured via Terraform.


onematchfox (approved):
LGTM. 

FYI: 
- fixed some indentations issues in the tests (mixture of tabs/spaces is painful)
- dropped the default as we can let ArgoCD handle this
- cherry-picked a couple fixes for flaky tests from #301

PR #325, opened 2023-08-04
Fix `ClientCertFile` when using local config

marcelobartsch:


oboukili (dismissed):

onematchfox:
@marcelobartsch thanks so much for taking the time to track this down and open this PR; sorry, for all the hassle! Also, great spot on the `ClientCertFile` :smile:. I'll cut a new release with these fixes ASAP.

onematchfox:
@marcelobartsch, now back in front of a PC and on closer examination, `opts.Context` is already being set [here](https://github.com/oboukili/terraform-provider-argocd/blob/a1d2840412b9a249c04a59a3631f2b4eaea6d9ce/internal/provider/model_provider.go#L197). So, possibly the only underlying issue was the client cert. I've dropped the change you made to set `opts.Context` (which also missed fetching the value from the relevant env var) from this PR. Would you be able to test again at some point and confirm if just the fix for `ClientCertFile` works for you?

marcelobartsch:
the change for clientCert file works!! and you were totally right on the context stuff, I had a extra env variable causing havoc in my setup.

Thanks you so much for the help on fixing this!!!

onematchfox (approved):

Issue #333, opened 2023-09-08
Add support for ApplicationSet level Application modification Policies


ahublersos:
### Description

ArgoCD 2.8 added support for [setting application modification policies at the ApplicationSet level](https://argo-cd.readthedocs.io/en/stable/operator-manual/applicationset/Controlling-Resource-Modification/#managed-applications-modification-policies). Previously these policies were set on the ApplicationSet controller and applied to all ApplicationSets.

### Potential Terraform Configuration

```hcl
resource "argocd_application_set" "guestbook" {
  metadata {
    name = "guestbook"
  }

  spec {
    generator {
      ...
     }
   }

  sync_policy { 
    preserve_resources_on_deletion = true 
    applications_sync = "create-update"
  }

  template {
    ...
  }
}
```

### References
[feat(appset): Policies create-only, create-update, sync per ApplicationSet](https://github.com/argoproj/argo-cd/pull/11462)

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

ahublersos:
Since this feature was only added in ArgoCD 2.8 I believe it requires a new feature flag and an update the ArgoCD version installed when running tests both locally and in CI.

PR #337, opened 2023-09-15
Add applications sync to application set sync policies

ahublersos:
Fixes https://github.com/oboukili/terraform-provider-argocd/issues/333.
Adds support for applications_sync to the application set resource.

I also nearly installed the test manifests into a live cluster by running `testacc_prepare_env` so I added a few safety safety nets to prevent that from happening in the future. Happy to split this out into a new PR if that's preferred.


ahublersos:
I don't know why the `TestAccArgoCDCluster_invalidSameServer` test is failing. That passed in local testing and doesn't seem related to changes in the application_set resource.

onematchfox:
> I don't know why the `TestAccArgoCDCluster_invalidSameServer` test is failing. That passed in local testing and doesn't seem related to changes in the application_set resource.

Yeah, that's just a flaky test. Everything passing now. Will try take a look at the changes a some point today/tomorrow morning.

argocd/structure_application_set.go
```diff
@@ -12,14 +12,14 @@ import (
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-func expandApplicationSet(d *schema.ResourceData, featureMultipleApplicationSourcesSupported bool) (metadata meta.ObjectMeta, spec application.ApplicationSetSpec, err error) {
+func expandApplicationSet(d *schema.ResourceData, featureMultipleApplicationSourcesSupported bool, featureApplicationSetApplicationsSyncPolicy bool) (metadata meta.ObjectMeta, spec application.ApplicationSetSpec, err error) {
```
onematchfox, line 15: I realise this is a bit confusing, but can you please drop this parameter (and it's usage in this file) and rather use the feature flag to implement a check on the user's provided configuration as per https://github.com/oboukili/terraform-provider-argocd/blob/d84d09101f8c1954cf613df69890a846540692b7/argocd/resource_argocd_application_set.go#L49-L51.

In essence, this is the pattern used by all other feature flags - we check the user's configuration of the Terraform resource and return an error if they try to configure something via Terraform that isn't supported by their version of ArgoCD (thereby preventing the configuration from being applied to the resulting CRUD call the provider makes). Multiple applications sources is the exception here in that we support versions that both have and do not have the feature and thus, need [branching logic](https://github.com/oboukili/terraform-provider-argocd/blob/d84d09101f8c1954cf613df69890a846540692b7/argocd/structure_application_set.go#L800-L807) in what we set on the CRUD call.

onematchfox (changes requested):
Sorry for the slow feedback. Hopefully just one small change - otherwise LGTM.

argocd/structure_application_set.go
ahublersos: Got it, thanks for the explanation! I've removed this parameter and added a feature flag check in the resource.

onematchfox:
This isn't stale. I just need to get around to looking at the failing tests.

onematchfox (approved):

onematchfox:
Sorry for the delay here @ahublersos and thanks for your patience!

onematchfox:
@ahublersos can/should I go ahead and cut a new release now?

ahublersos:
@onematchfox yes, please do!

onematchfox:
> @onematchfox yes, please do!

Sorry... more delays. Our release pipeline is broken. Will try upgrade/fix this weekend.

PR #399, opened 2024-06-12
fix(argocd_cluster): use cluster list api to avoid 403 issues with cluster get api at cluster read time

w4rgrum:
This PR fixes https://github.com/argoproj-labs/terraform-provider-argocd/issues/266

Coded it with below comment in mind, meaning it is backward compatible: https://github.com/argoproj-labs/terraform-provider-argocd/issues/266#issuecomment-1739138091
* code does a `Get` as usual first (legacy code)
* if it fails with a `PermissionDenied` error (i.e. `403`) the fix is triggered (in a backward compatible way for servers <`v2.8`, see below)
   In that case a call is issued to the `List` api, reusing some code used for create cluster existence checks (that already uses this `List` api). Noe that this code is able to filter the cluster in the resulting list if not yet done by the ArgoCD server (servers filtering on `List` api does not work if <`v2.8`, see comments in linked issue and/or in the code ccomments)

Also added a non-regression test to validate a server-side deletion of the cluster triggers the fix.


w4rgrum:
@onematchfox PR is ready to be reviewed :)

morhekil:
any chance to have this released as new version of the provider? This seems to be a blocker when using the provider with recent ArgoCD versions

onematchfox:
...

onematchfox:
...

onematchfox:
...

onematchfox:
...

onematchfox:
.

argocd/resource_argocd_cluster.go
```diff
@@ -53,6 +56,7 @@ func resourceArgoCDClusterCreate(ctx context.Context, d *schema.ResourceData, me
 		return errorToDiagnostics(fmt.Sprintf("failed to list existing clusters when creating cluster %s", cluster.Server), err)
 	}
 
+	// Here we will filter ourselves on the list so that we are backward compatible for argo-cd server with version < v2.8.0 (see coment above)
```
mkilchhofer, line 59: Do we need backward compat to versions below 2.8? That is outside of our promise, no?

the-technat (approved):
LGTM, I like the way it was implemented for backwards-compatibility, although I don't think we need to implement backwards-compatibility to unsupported Argo CD versions in other places, it's already there so we should give it a go ;).

argocd/resource_argocd_cluster.go
the-technat: As mentioned In my review I don' think we need that compat, but since it's already coded in a way that doesn't harm we should leave it as is so far.

mkilchhofer:
@onematchfox anything needed from your side?

the-technat:
Thanks @w4rgrum  for your patience and contributing this! 

We will soon release a new version with this fix, but want to wait for #472 before doing so.

PR #411, opened 2024-08-06
feat: add support for `validate` option when provisioning/maintaining argocd applications

jiachengxu:
resolves https://github.com/argoproj-labs/terraform-provider-argocd/issues/261
this PR adds `validate` (similar to `--validate` flag in argocd CLI) to `argocd_application` resource, to be able to skip(set it to false) the app validation. It is set to true by default.


onematchfox (approved):
Thanks for picking this one up. LGTM.

PR #420, opened 2024-10-01
docs: Move Contributing docs to separate file

mkilchhofer:
Motivation:
- match with all other Argoproj / Argoproj-labs repos
- (CLOmonitor checks for that in case we activate it sometimes in the future)


the-technat (approved):
Given your arguments I think that makes absolutely sence.

PR #444, opened 2024-10-04
feat: update used go version to 1.22

the-technat:
Inspired by this issue: https://github.com/hashicorp/terraform-provider-random/issues/377


the-technat:
will be reopened at a later time...

@mkilchhofer already bumped to 1.21 and till we can go further we have to bump a lot of other dependencies.

PR #450, opened 2024-10-07
ci: Don't always execute the long-running acceptance tests

mkilchhofer:
We trigger CI way too much. Let's optimize it:
- Don't trigger on main branch (after a PR got merged)
  **Reason:** We enforce linear commits (contributors and maintainers need to have current CI run on the PR).
- Don't trigger CI when Dependabot updates other workflows
  > image
- Don't trigger on README and CONTRIBUTING changes

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -2,10 +2,14 @@
 name: Tests
 
 on:
-  push:
-    branches: [main]
   pull_request:
-    branches: [main]
+    branches:
+    - main
+    paths:
+    - "!README.md"
+    - "!CONTRIBUTING.md"
+    - "!.github/**"
+    - ".github/workflows/tests.yml"
 
 jobs:
   build:


onematchfox:
> * **Reason:** We enforce linear commits (contributors and maintainers need to have current CI run on the PR).

Is this something you have enabled recently? Might just be my poor memory (:sweat_smile:) but I don't remember this being the case previously.

mkilchhofer:
> Is this something you have enabled recently? Might just be my poor memory ( :sweat_smile: ) but I don't remember this being the case previously.

I changed some minor bits but I cannot remember that I changed this. Also my statement was not 100% correct, we only allow the option "Allow squash merging" but the other regarding linear history is not active. But still we as maintainers cannot click the merge button when the source branch in the PR is not up-to-date:
> image

the-technat (approved):

onematchfox:
> I changed some minor bits but I cannot remember that I changed this. Also my statement was not 100% correct, we only allow the option "Allow squash merging" but the other regarding linear history is not active.

This is because require branches to be up to date before merging is enabled in the  branch protection on `main`. 

The reason I ask is that whilst I totally agree that this is good practice, I wonder if it's not going to cause friction in the review process. If maintainers do not have permission to push to PR branches then anytime a PR is merged, every other author with an open PR will once again need to align with `main` won't they?

mkilchhofer:
> This is because require branches to be up to date before merging is enabled in the branch protection on `main`.

Ahh yep, this option is checked in branch protection:
> **Require branches to be up to date before merging**
> This ensures pull requests targeting a matching branch have been tested with the latest code. This setting will not take effect unless at least one status check is enabled (see below).


> The reason I ask is that whilst I totally agree that this is good practice, I wonder if it's not going to cause friction in the review process. If maintainers do not have permission to push to PR branches then anytime a PR is merged, every other author with an open PR will once again need to align with `main` won't they?

This can happen when contributors file a PR from a repo livin in an organization (we've seen this 1-3 times in argo-helm). But normally on personal contributions, we can ask the contributor to check the checkbox on the PR, no big deal :)

mkilchhofer:
@onematchfox maybe a compromise is to only run it as-is on the pull-request but not on the main branch?

mkilchhofer:
Let's close it for now and optimize at a later time :)

onematchfox:
Sorry, this had slipped my mind. If we're going to enforce branches to be up to date before merging then sure, there's no need to run the build after merging. I was just caught unaware by the change to enforce that branches are up to date - that's a bigger change to workflow.

Issue #445, opened 2024-10-04
bump argocd go dep to 2.10


KyriosGN0:
### Potential Terraform Configuration

```hcl
# Copy-paste your Terraform configurations here - for large Terraform configs,
# please use a service like Dropbox and share a link to the ZIP file. For
# security, you can also encrypt the files using our GPG public key.
```

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

mkilchhofer:
@the-technat is working on that:
- https://github.com/argoproj-labs/terraform-provider-argocd/pull/443

PR #453, opened 2024-10-08
feat: replace protobuf with new module

the-technat:
This replaces `github.com/golang/protobuf` with it's superseeded `google.golang.org/protobuf` module.

See [github.com/golang/protobuf's README](https://github.com/golang/protobuf) for the notice that this has moved to a new location.

Will hopefully make it easier to work on #445 as there are a lot of changes we need to reflect before we can bump the argo-cd module.


mkilchhofer (approved):

onematchfox (approved):

blakepettersson (approved):
LGTM

PR #455, opened 2024-10-10
chore: upgrade k8s modules to suport v1.29

asdfgugus:


the-technat (commented):
The linter config has to be updated if we bump to go 1.22 (see failed test).

But anyway, can we wait with this til #443 is merged? I'm not sure  how these PR's relate to each other, maybe they need to go together for successful builds.

asdfgugus:
Sounds good. Anyways, I have to invest more time to upgrade the modules. If you find the time, feel free to upgrade it yourself.

mkilchhofer:
Maybe we can include update-script (hack) from the upstream argo-cd repo:

Xref:
- https://github.com/argoproj/argo-cd/pull/16581

mkilchhofer:
Already implemented within #470

PR #522, opened 2024-12-18
feat: Extend policy validation to support update and delete actions

### LINKED ISSUE #524: Unable to Define Policy with Specific Actions in ArgoCD Provider
Description:  
Using the provider, it is not possible to define a policy with actions such as:  
- `delete/*/Pod/*/*`  
- `delete/*/ReplicaSet/*/*`  

However, according to the documentation, it should be possible to define actions in this format.  

```
Terraform version: Terraform v1.9.6
ArgoCD provider version: 7.1.0
ArgoCD version: v2.13.1
```

### Affected Resource(s)

- argocd_project

### Expected Behavior

The provider should allow defining policies with actions as described in the [documentation](https://argo-cd.readthedocs.io/en/stable/operator-manual/rbac/#fine-grained-permissions-for-updatedelete-action).

### Actual Behavior

The actions mentioned above cannot be defined using the provider.

### Community Note

pouriyabp:
This fix enhances the policy validation function to support the update and delete actions, in addition to the previously supported action keyword.

Per the documentation on [Fine-grained Permissions for update/delete actions](https://argo-cd.readthedocs.io/en/stable/operator-manual/rbac/#fine-grained-permissions-for-updatedelete-action), the update and delete actions allow users to perform operations on an application and all of its resources. Additionally, these actions can be scoped to specific resources using the ```<action>/<group>/<kind>/<ns>/<name>``` format.

To align with this functionality, the regex used in policy validation has been updated to recognize update and delete as valid actions. This ensures that policies leveraging fine-grained permissions for update and delete are now properly validated.

---

Fixes #524


pouriyabp:
[issue](https://github.com/argoproj-labs/terraform-provider-argocd/issues/524)

mkilchhofer:
Xref:
- https://github.com/argoproj/argo-cd/pull/19338

argocd/utils.go
```diff
@@ -97,7 +97,7 @@ func isValidPolicyAction(action string) bool {
 		"*":                       true,
 	}
 	validActionPatterns := []*regexp.Regexp{
-		regexp.MustCompile("action/.*"),
+		regexp.MustCompile("(action|update|delete)/.*"),
```
mkilchhofer, line 100: Can we use the form which is available in upstream Argo CD?
```suggestion
		regexp.MustCompile("action/.*"),
		regexp.MustCompile("update/.*"),
		regexp.MustCompile("delete/.*"),
```

See PR https://github.com/argoproj/argo-cd/pull/19338/files

grafik

mkilchhofer (changes requested):
Hi @pouriyabp sorry for the late response. Only a small finding, then I think we are good.

argocd/utils.go
pouriyabp: Hi @mkilchhofer, no problem at all, and thanks for the review!

mkilchhofer:
@the-technat @onematchfox do we need a conditional with a new feature toggle/gate as these two new regexes are only valid in Argo CD 2.12+?

I assume that Argo CD <2.12 will decline the creation of projects with these regexes and thus the provider will throw the error to the user?

mkilchhofer:
@pouriyabp I added a test for you so our CI will validate your enhancement of the regexes.

the-technat:
Maybe we can add a version validation here: https://github.com/argoproj-labs/terraform-provider-argocd/blob/main/internal/features/features.go

Other features already use this to check if it's supported, just not sure how difficult it is to implement for a regex extension.

the-technat:
Maybe we can add a version validation here: https://github.com/argoproj-labs/terraform-provider-argocd/blob/main/internal/features/features.go

Other features already use this to check if it's supported, just not sure how difficult it is to implement for a regex extension.

mkilchhofer:
> Maybe we can add a version validation here: https://github.com/argoproj-labs/terraform-provider-argocd/blob/main/internal/features/features.go
> 
> Other features already use this to check if it's supported, just not sure how difficult it is to implement for a regex extension.

I added the feature conditional on the test only:
image
image

And to check it already on the regex, we need to pass the Argo server version down to utils. Do we want that, @the-technat?

the-technat (approved):

mkilchhofer (approved):

PR #360, opened 2023-12-26
Add support for Kustomize patches to `argocd_application`

chidambaram27:
Closes #356

PR #530, opened 2025-01-08
Add support for Kustomize patches to argocd_application - alternative

the-technat:
Alternative to #360 with changes from Brian (@onematchfox)


the-technat:
The changes from this PR to #360  have been merged into #360, so this draft is no longer needed.

PR #531, opened 2025-01-08
ci: install terraform before running generate

the-technat:
Fixes the CI tests which are currently failing.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -29,6 +29,12 @@ jobs:
       with:
         version: v1.59.1
 
+    # We need the latest version of Terraform for our documentation generation to use
+    - uses: hashicorp/setup-terraform@b9cd54a3c349d3f38e8881555d616ced269862dd # v3.1.2
+      name: Setup Terraform
+      with:
+        terraform_wrapper: false
+
     - name: Generate
       run: make generate
 


mkilchhofer (approved):

onematchfox (approved):
How has this worked until now? :astonished:

the-technat:
@onematchfox good question, probably the github runners had a Terraform binary installed by default.

Anyway, the scaffold repo does this too: https://github.com/hashicorp/terraform-provider-scaffolding-framework/blob/main/.github/workflows/test.yml#L45-L48

onematchfox:
Complete nit... Not that this repo requires/enforces commit message semantics to begin with but, for future reference, `ci` would be a more appropriate commit type than `feat(ci)` for this change since it isn't actually adding anything to the provider itself. I.e. `build: install terraform before running generate`. See [angular/CONTRIBUTING.md](https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#type) for reference on the meaning of commit types :smile:.

mkilchhofer:
@onematchfox @the-technat I found out what the root cause was: https://github.blog/changelog/2024-11-05-notice-of-breaking-changes-for-github-actions/

Terraform was included in Ubuntu 22 while it is not included anymore in Ubuntu 24:
- https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md
- https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md

PR #566, opened 2025-03-11
feat: use pinned terraform versions in testing matrix

the-technat:
Currently we use the "latest" terraform binary available in the Github runner. This is unpredictable and causes tests to suddenly be unstable if Terraform changes something from one minor version to another (as they likely did from 1.3 to 1.4, read https://github.com/runatlantis/atlantis/issues/2242).

Currently we only test one Terraform version. This change keeps this behavior but allows for multiple version testing in the future, in case we want this.


mkilchhofer (approved):
LGTM. lets hope that this fixes our flaky CI :) then we can bump this version and maybe we find out what TF version breaks our CI

PR #567, opened 2025-03-11
Feat: Add support for destination service accounts

nimjor:


mkilchhofer:
Xref:
- https://argo-cd.readthedocs.io/en/stable/operator-manual/app-sync-using-impersonation/#configuring-destination-service-accounts
- https://github.com/argoproj/argo-cd/pull/17403
- https://github.com/argoproj/argo-cd/issues/20336

internal/features/features.go
```diff
@@ -17,26 +18,31 @@ const (
...
-	ApplicationSet:                             {"application sets", semver.MustParse("2.5.0")},
-	ApplicationSetProgressiveSync:              {"progressive sync (`strategy`)", semver.MustParse("2.6.0")},
-	ManagedNamespaceMetadata:                   {"managed namespace metadsata", semver.MustParse("2.6.0")},
-	ApplicationSetApplicationsSyncPolicy:       {"application set level application sync policy", semver.MustParse("2.8.0")},
-	ApplicationSetIgnoreApplicationDifferences: {"application set ignore application differences", semver.MustParse("2.9.0")},
-	ApplicationSetTemplatePatch:                {"application set template patch", semver.MustParse("2.10.0")},
-	ApplicationKustomizePatches:                {"application kustomize patches", semver.MustParse("2.9.0")},
-	ProjectFineGrainedPolicy:                   {"fine-grained policy in project", semver.MustParse("2.12.0")},
-	ApplicationSourceName:                      {"named application sources", semver.MustParse("2.14.0")},
```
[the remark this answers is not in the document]
nimjor: Definitely not intentional, thanks for catching that--probably a bad merge conflict resolve.  I've added it back in.

internal/provider/server_interface.go
```diff
@@ -159,12 +167,60 @@ func (si *ServerInterface) InitClients(ctx context.Context) diag.Diagnostics {
...
+			tflog.Error(ctx, fmt.Sprintf("error evaluating settings check expression '%v': %v", rs, err))
+			return false
+		} else if result == nil || !result.(bool) {
+			tflog.Debug(ctx, fmt.Sprintf("settings check expression '%v' evaluated to false", rs))
+			return false
+		}
+	}
+
+	return true
+}
```
onematchfox, line 208: Whilst I appreciate the thought behind the addition of this function, it isn't ever used at present. The newly added `FeatureConstraint` is only ever used in the tests and thus, is only checked by `testAccPreCheckFeatureSupported` which, simply [checks the `MinVersion`](https://github.com/argoproj-labs/terraform-provider-argocd/blob/777d110ffa0e85b8c7ebfc488fa6945de33e57fd/argocd/provider_test.go#L93-L95). Now, whilst we could address this within the PR, I can't help but wonder whether this check is necessary at all. It's addition adds a lot of complexity - is that complexity worth it? 

The existing feature checks we have (for `MinVersion`) are about preventing the provider from failing if it makes API calls containing configuration that is not supported by the user's ArgoCD version (i.e. when they try to make use of a field/configuration that doesn't yet exist in the version of ArgoCD that they are running). However, in this case, it is completely possible to create an `AppProject` with `destinationServiceAccounts` set without enabling the "application sync with impersonation" feature in ArgoCD. The provider won't fail if an alpha feature is not enabled. So, the question is, should the provider even care about whether or not ArgoCD is configured to do anything with the configuration if the configuration applies correctly?

My 2 cents (happy to hear other opinions) is that it shouldn't, and, as such, we should remove this check. Whilst I value the idea of shifting left and informing the user that their intended configuration won't have the desired effect, I don't think that's worth the additional complexity. ArgoCD itself doesn't even do this - you can happily create a project via the CLI with destination service accounts configured even if the feature flag is disabled (without any warning/error). Why should the provider be any different? Hence, IMO, we should ditch this[^1] and simply refer the user to the [ArgoCD documentation](https://argo-cd.readthedocs.io/en/release-2.14/operator-manual/app-sync-using-impersonation) from within the provider docs. 

``` bash
$ kubectl get cm argocd-cm -o yaml | grep impersonation
$ argocd proj create foo --dest-service-accounts https://192.168.99.100:8443,default,default-sa
$ argocd proj list                                                                             
NAME        DESCRIPTION  DESTINATIONS                      SOURCES  CLUSTER-RESOURCE-WHITELIST  NAMESPACE-RESOURCE-BLACKLIST  SIGNATURE-KEYS  ORPHANED-RESOURCES    DESTINATION-SERVICE-ACCOUNTS
default                  *,*                               *        */*                         <none>                        <none>          disabled              <none>
foo                      <none>                            <none>   <none>                      <none>                        <none>          disabled              https://192.168.99.100:8443,default,default-sa
``` 

[^1]: Removing "this" function also entails undoing changes to propagate context to `IsFeatureSupported` as well as the introduction of  `INCL_UNSTABLE_FEATURES` and related changes to CI/test workflows to use. Can also simplify the newly added test `TestAccUnstableArgoCDProjectWithDestinationServiceAccounts` as we no longer need to rely on the `kubernetes` provider to apply additional configuration.

argocd/resource_argocd_project.go
```diff
@@ -164,7 +164,7 @@ func resourceArgoCDProjectUpdate(ctx context.Context, d *schema.ResourceData, me
 		return errorToDiagnostics(fmt.Sprintf("failed to expand project %s", d.Id()), err)
 	}
```
onematchfox, line 165:
```sugggestion
	}

	if !si.IsFeatureSupported(ctx, features.ProjectDestinationServiceAccounts) {
		_, destinationServiceAccountsOk := d.GetOk("spec.0.destination_service_account")
		if destinationServiceAccountsOk {
			return featureNotSupported(features.ProjectDestinationServiceAccounts)
		}
	}
```

Please check if the feature is enabled during creation/update if the configuration is used.

argocd/resource_argocd_project.go
```diff
@@ -164,7 +164,7 @@ func resourceArgoCDProjectUpdate(ctx context.Context, d *schema.ResourceData, me
...
 
-	if !si.IsFeatureSupported(features.ProjectSourceNamespaces) {
+	if !si.IsFeatureSupported(ctx, features.ProjectSourceNamespaces) {
 		_, sourceNamespacesOk := d.GetOk("spec.0.source_namespaces")
 		if sourceNamespacesOk {
 			return featureNotSupported(features.ProjectSourceNamespaces)
```
onematchfox, line 170:
```sugggestion
			return featureNotSupported(features.ProjectSourceNamespaces)
		}
	}

	if !si.IsFeatureSupported(ctx, features.ProjectDestinationServiceAccounts) {
		_, destinationServiceAccountsOk := d.GetOk("spec.0.destination_service_account")
		if destinationServiceAccountsOk {
			return featureNotSupported(features.ProjectDestinationServiceAccounts)
		}
	}
```

onematchfox (changes requested):

the-technat:
@nimjor we just merged a refactoring PR (#685) that migrated some resources to the terraform-plugin-framework and thus had to move the server_interface back to the SKDv2 code. I'm afraid that means some of your code has to be moved as well 🙈.

internal/provider/server_interface.go
the-technat: Interesting thoughts @onematchfox. I like the idea, given that we do the same in a lot of other places (api client options for example). The only thing I'm concerned about is whether we are still able to test this properly?

onematchfox: > The only thing I'm concerned about is whether we are still able to test this properly?

What are you referring to by "this"? What are we trying to test from the provider's perspective? If you're referring to whether or not the specified configuration (in this case, "destination service accounts") is successfully _maintained_ by the provider, then my answer would be yes, we are still able to test this properly; the tests we have ensure that the configuration can be both applied and read back from ArgoCD.

nimjor: Since it sounds like we're basically all on the same page, I'll go ahead and undo the IsSettingSupported changes.

nimjor:
> @nimjor we just merged a refactoring PR (#685) that migrated some resources to the terraform-plugin-framework and thus had to move the server_interface back to the SKDv2 code. I'm afraid that means some of your code has to be moved as well 🙈.

@the-technat let me know if there's anything specific in this PR in the current state that needs to get moved.

onematchfox (approved):
Thanks @nimjor!

the-technat (approved):
Thanks @nimjor  for your patience!

PR #589, opened 2025-03-14
feat: add compat with Argo CD 2.14

### LINKED ISSUE #581: Add support for Argo CD 2.14
### Description

Now that Argo CD 2.14 is out we should ensure we use the latest client & test against it.

the-technat:
Imports the argocd client for 2.14 ensuring optimal compatibility with Argo CD 2.14 and tests against Argo CD 2.14.

Related to that we also ensure we use the same client-go version as Argo CD uses.

Closes #581


mkilchhofer (dismissed):
Looks good :)

internal/provider/model_provider.go
```diff
@@ -124,7 +123,7 @@ func (p ArgoCDProviderConfig) getApiClientOptions(ctx context.Context) (*apiclie
 			runtime.ErrorHandlers = runtimeErrorHandlers
 		}
 
-		err := headless.MaybeStartLocalServer(ctx, opts, "", nil, nil, cache.RedisCompressionNone, nil)
+		err := headless.MaybeStartLocalServer(ctx, opts, "", nil, nil, &clientcmd.DirectClientConfig{})
```
mkilchhofer, line 126: https://github.com/argoproj/argo-cd/pull/21901/files#diff-170ef34ea6636f7c5461828e058b61f2e4e53f6f90d761e07be2fa1637d9eaf6
maybe `nil` instead of ` &clientcmd.DirectClientConfig{}`

the-technat: changed, now it's `nil`

onematchfox (approved):

mkilchhofer (approved):

PR #607, opened 2025-03-26
docs: add explanation about our dependency management + clarify version compatiblity in README

### LINKED ISSUE #440: Clarify (and improve) testing reliability
We should have a look into multiple parts of testing:
- Acceptance tests sometimes fail (relatively often)
- Against what versions of Argo CD should we test (currently we test against EOL versions of Argo CD)
- [internal/provider/data_source_application_test.go](https://github.com/argoproj-labs/terraform-provider-argocd/blob/main/internal/provider/data_source_application_test.go)  why do we specify a version constraint here? Can we remove it?
- ..? (list not complete)

the-technat:
I think the compatibility promise needs to be clarified, thus I reworded it.

In the contributor docs I tried to write down our handling of dependencies, so that it's obvious for contributors why we use which version of specific dependencies.

Oh and since the provider-move to a new location is now a few months old I moved the upgrade section down a bit in the README.

Closes #440 and #612

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,6 +2,21 @@
 
 Contributions are welcome! 
 
+## Dependency Management
+
+### K8s version
+In our CI we test against a Kubernetes version that is supported by all Argo CD versions we support.
+
+That version can be obtained when looking at [this table](https://argo-cd.readthedocs.io/en/stable/operator-manual/installation/#tested-versions) in the Argo CD documentation.
+
+### Argo CD client-lib
+
+Some dependencies we use are strictly aligned with the Argo CD client-lib that we use and should only be updated together:
+- github.com/argoproj/gitops-engine
+- k8s.io/*
+
+Please don't update any of these dependencies without having discussed this first!
+
 ## Building
 
 1. `git clone` this repository and `cd` into its directory
--- a/README.md
+++ b/README.md
@@ -33,15 +33,16 @@ Registry](https://registry.terraform.io/providers/argoproj-labs/argocd/latest/do
 
 The provider supports all versions Argo-CD itself currently supports. See the [Argo CD documentation](https://argo-cd.readthedocs.io/en/stable/operator-manual/installation/#supported-versions) for supported versions.
 
-The following table shows which provider version to use given your Argo CD version:
+The following table shows the versions of Argo CD that are tested with each version of the provider.
 
-| Argo CD version | minimum provider version | maximum provider version |
-|-----------------|--------------------------|--------------------------|
-| 2.14            | 7.5.0                    | not yet available        |
-| 2.13            | 7.4.0                    | not yet available        |
-| 2.12            | 7.2.0                    | not yet available        |
-| 2.11            | 7.1.0                    | 7.4.0                    |
-| 2.10            | 6.1.0                    | 7.3.1                    |
+| Provider version | Argo CD versions |
+|---|---|
+| 7.5 | v2.12, v2.13, v2.14 | 
+| 7.4 | v2.11, v2.12, v2.13 | 
+| 7.2 | v2.10, v2.11, v2.12 | 
+| 7.1 | v2.9, v2.10, v2.11 | 
+| 7.0 | v2.8, v2.9, v2.10 | 
+| 6.x | v2.8, v2.9, v2.10 | 
 
 Note: these numbers are based on our testing matrix that tests the provider against these versions of Argo CD. You may be able to use other provider versions as the ones listed here and it may still work.
 
--- a/README.md
+++ b/README.md
@@ -29,6 +29,51 @@ Official documentation on how to use this provider can be found on the
 [Terraform
 Registry](https://registry.terraform.io/providers/argoproj-labs/argocd/latest/docs).
 
+## Version Compatibility 
+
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -29,6 +29,41 @@ Official documentation on how to use this provider can be found on the
 [Terraform
 Registry](https://registry.terraform.io/providers/argoproj-labs/argocd/latest/docs).
 
+## Version Compatibility 
+
+We aim to support `n-2` versions of Argo CD, where `n` is the latest available minor version. Currently this means we support Argo CD `2.14`, `2.13` and `2.12`.
```
onematchfox, line 34: Maybe we should simplify this to something like we support current active version of ArgoCD and link to their docs? Explicitly calling out the current version is going to be a maintenance burden?

CONTRIBUTING.md
```diff
@@ -2,6 +2,17 @@
 
 Contributions are welcome! 
 
+## Dependency Management
+
+The README states that we support `n-2` Argo CD versions, where `n` is the current minor version of Argo CD. 
+
+To do so we always import the exact Argo CD client-lib we provide support for.
```
onematchfox, line 9: Not sure I follow. If we support `n-2` versions how can we import "the exact" client-lib we provide support for?

the-technat: I see that's a bit misleading. I meant if we support `2.14`, `2.13` and `2.12` we import the `2.14` client-lib. 

But this is a bit odd as well.

README.md
onematchfox: Yeah, exactly. We support all currently active version of ArgoCD.

CONTRIBUTING.md
onematchfox: Yeah, so historically we only bumped the client lib dependency when needed (security patches, new features being adopted) as this way we maintain compatability with older version for as long as possible.

README.md
the-technat: I reworded the section, WDYT now?

CONTRIBUTING.md
onematchfox: > But how then do you ensure it's working with the current versions of Argo CD that are supported upstream?

That is ensured by updating the versions of ArgoCD we test against in the test matrix in the GitHub action.

the-technat: All right, do you have a suggestion how to write this down in the contributor docs?

README.md
```diff
@@ -29,6 +29,51 @@ Official documentation on how to use this provider can be found on the
...
+
+The following table shows which provider version to use given your Argo CD version:
+
+| Argo CD version | minimum provider version | maximum provider version |
+|-----------------|--------------------------|--------------------------|
+| 2.14            | 7.5.0                    | not yet available        |
+| 2.13            | 7.4.0                    | not yet available        |
+| 2.12            | 7.2.0                    | not yet available        |
+| 2.11            | 7.1.0                    | 7.4.0                    |
+| 2.10            | 6.1.0                    | 7.3.1                    |
```
onematchfox, line 44:
```suggestion
The following table shows the versions of Argo CD that are tested with each version of the provider.

| Provider version | Argo CD versions |
|---|---|
| 7.5 | v2.12, v2.13, v2.14 | 
| 7.4 | v2.11, v2.12, v2.13 | 
| 7.2 | v2.10, v2.11, v2.12 | 
| 7.1 | v2.9, v2.10, v2.11 | 
| 7.0 | v2.8, v2.9, v2.10 | 
| 6.x | v2.8, v2.9, v2.10 | 
```

I'd take inspiration from [Argo CD itself](https://argo-cd.readthedocs.io/en/stable/operator-manual/installation/#tested-versions) and simplify this table to just state the known working/tested versions of Argo CD. We (or at least I :smile:) don't want to be tracking max/minimum working versions of Argo CD per provider version in perpetuity. We should probably also think about how long we maintain a list of older provider versions and/or whether the table needs to be updated for every minor release of the provider vs only when tested versions change.

mkilchhofer: But this way have a lot longer list (duplicated lines) as we also might bump the minor provider version for just newly added features within one Argo CD version. A fictive example:

- Provider version 25.1(.x) added Argo CD 2.1 support
- Provider version 25.2.(x) adds support for feature X within Argo CD 2.1

That would be much more work than just writing the last 3-5 Argo CD versions to a list and writing min/max versions into a list.

But of course we can overengineer it and create that list (that might contain dups) via CI ;) 😆

onematchfox: Agreed. That is why I said

> We should probably also think about how long we maintain a list of older provider versions and/or whether the table needs to be updated for every minor release of the provider vs only when tested versions change.

If you look at my suggestion I intentionally excluded `7.3` since that didn't change the tested versions.

onematchfox: > there are no duplicated lines

OK, so the tested versions are duplicated between `6.x` and `7.0`. Question is at this point, how far back we want to go to start with and then, going forward, how the table is updated and how often old versions are pruned.

the-technat: That's where I think a support policy of some kind might make sense. Eiher for the versions of the provider or versions of Argo CD. Cuase then it's obvious when to prune old versions?

onematchfox (approved):

mkilchhofer (approved):

PR #626, opened 2025-04-11
fix: allow setting of `applicationsets` in appproject policies

blakepettersson:
Since argoproj/argo-cd#21829 it's been possible to set `applicationsets` and other resources as a project-role policy, but it turns out that this provider also does its own validation.

To fix this, we upgrade Argo CD to 2.14.9 and add the relevant resources to the `validResources` map.


the-technat (approved):
LGTM

mkilchhofer (approved):

blakepettersson:
Thanks guys!

PR #634, opened 2025-04-29
fix: allow templated label values

blakepettersson:
`validateMetadataLabels` ensures that labels are in a certain format. This works fine in most cases, except when this is used within appsets, which allow for labels to be templated. This adds a layer of indirection allowing for templated labels to be used, but only when an appset makes use of this feature.


the-technat (approved):
Nice fix!

mkilchhofer (approved):

PR #624, opened 2025-04-09
feat: add support for appset plugin generators

blakepettersson:
This has one caveat, which likely cannot be fixed until #615 is merged; we're limited to simple string maps for `parameters`, which is only a subset of what the plugin generator supports. In reality pretty much anything can go in as a parameter from the perspective of the plugin generator, but `terraform-plugin-sdk` needs an explicit type.

With `terraform-plugin-framework` we can use any as a type, then we can handle the marshaling/unmarshaling appropriately. I don't believe this will be a breaking change in the future since we will be widening the types `parameters` will accept at a later stage.

Fixes #620

PR #639, opened 2025-05-05
fix: add plugin generator to nested generator

blakepettersson:
#624 added support for plugin generators, but the support for it was never added to the nested generators.


blakepettersson:
Drafting until tests are added for this PR

the-technat (approved):

onematchfox (approved):

PR #641, opened 2025-05-06
feat: ensure support with Argo CD v3

the-technat:
Ensuring we are up-to-date with dependencies and support Argo CD V3.


onematchfox:
Out of interest sake, what happens if you only upgrade the test matrix and not the client? Is `7.6` compatible with `3.0` out the box or are there changes that require that we bump the client library version?

the-technat:
> Out of interest sake, what happens if you only upgrade the test matrix and not the client? Is `7.6` compatible with `3.0` out the box or are there changes that require that we bump the client library version?

I tried running the acceptance tests locally and they seem to pass with the current provider version.

mkilchhofer (approved):

Issue #548, opened 2025-02-10
Support ApplicationSets in any Namespace


VerhovnikU:
We are encountering an error when attempting to deploy an ApplicationSet resource into a namespace different from the namespace where our ArgoCD instance is deployed. When running terraform apply, Terraform reports an error similar to the following:

```

Error: Provider produced inconsistent result after apply

When applying changes to module.argocd[0].argocd_application_set.this["appset"], provider "registry.terraform.io/argoproj-labs/argocd" produced an unexpected new value: Root object was present, but now absent.

```

Additionally, our check results show that while the variable for application_sets passes validation

```
"check_results": [
    {
      "object_kind": "var",
      "config_addr": "module.argocd.var.application_sets",
      "status": "pass",
      "objects": [
        {
          "object_addr": "module.argocd[0].var.application_sets",
          "status": "pass"
        }
      ]
    }
  ]
```

Reproduction Steps:

Configure an argocd_application_set resource using the provider, explicitly setting the metadata namespace to one that is different from the namespace where ArgoCD is deployed. For example:

```
resource "argocd_application_set" "gitlab-runners" {
  metadata {
    name      = "app"
    namespace = "test"  // namespace different from ArgoCD's namespace
  }

  spec {
    generator {
      // your generator configuration
    }

    template {
      metadata {
        name = "app-{{name}}"
      }
      spec {
        source {
          repo_url        = "https://github.com/your/repo.git"
          target_revision = "HEAD"
          path            = "path/to/app"
        }
        destination {
          server    = "{{server}}"
          namespace = "dest-namespace" 
        }
      }
    }
  }
}

```
Additional Context:

This issue appears to be triggered by cross‑namespace deployments. When the ApplicationSet is deployed in the same namespace as the ArgoCD instance, the error does not occur.
The ApplicationSet is deployed but terraform state doesnt recognise it. We have Cluster Roles defined for AppSet in any namespace

Request:
Could the maintainers please investigate whether this behavior is a bug in the provider or a configuration nuance that requires additional documentation? Any insights or workarounds would be greatly appreciated.

VerhovnikU:
Any comments ?

the-technat:
I can reproduce the issue, here's a full-reproduce code snippet:

```hcl
# On a fresh K8s cluster intall Argo CD using:
[...]

PR #644, opened 2025-05-09
feat(appsets): add state upgrade logic for appsets in any namespace

the-technat:
Adds a state migration to automatically update existing resources to use the new ID format. This should complete #548 and allow for a non-breaking change.

Most of the code was copied from #212

### DIFF
--- a/argocd/schema_application_set.go
+++ b/argocd/schema_application_set.go
@@ -140,12 +140,11 @@ func applicationSetSpecSchemaV0() *schema.Schema {
 }
 
 func applicationSetSpecSchemaV1() *schema.Schema {
-	// To support deploying applicationsets to non-default namespaces (aka project
-	// source namespaces), we need to do a state migration to ensure that the Id
-	// on existing resources is updated to include the namespace.
-	// For this to happen, we need to trigger a schema version upgrade on the
-	// application resource however, the schema of the application `spec` has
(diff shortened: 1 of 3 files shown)


argocd/schema_application_set.go
```diff
@@ -136,6 +139,16 @@ func applicationSetSpecSchemaV0() *schema.Schema {
 	}
 }
 
+func applicationSetSpecSchemaV1() *schema.Schema {
+	// To support deploying applicationsets to non-default namespaces (aka project
+	// source namespaces), we need to do a state migration to ensure that the Id
+	// on existing resources is updated to include the namespace.
+	// For this to happen, we need to trigger a schema version upgrade on the
+	// application resource however, the schema of the application `spec` has
+	// changed from `v0`.
```
onematchfox, line 148:
``` suggestion
	// To support deploying applicationsets to non-default namespaces we need to
	// do a state migration to ensure that the Id on existing resources is
	// updated to include the namespace. For this to happen, we need to trigger
	// a schema version upgrade on the applicationset resource however, the
	// schema of the applicationset `spec` has not changed from `v0`.
``` 


nit: slight copy/pasta

onematchfox (approved):
LGTM

mkilchhofer (approved):

PR #651, opened 2025-05-18
Feat: Support for Application Source Name

jodhi:
Hello,
This PR related to https://github.com/argoproj/argo-cd/pull/20470 to add name field in the application source object.
Impacted component will be `argocd_application` resource & data source.

Let me know if I missed something. thanks

### DIFF
--- a/argocd/resource_argocd_application_test.go
+++ b/argocd/resource_argocd_application_test.go
@@ -1062,7 +1062,11 @@ func TestAccArgoCDApplication_CustomNamespace(t *testing.T) {
 
 func TestAccArgoCDApplication_MultipleSources(t *testing.T) {
 	resource.ParallelTest(t, resource.TestCase{
-		PreCheck:          func() { testAccPreCheck(t); testAccPreCheckFeatureSupported(t, features.MultipleApplicationSources) },
+		PreCheck: func() {
+			testAccPreCheck(t)
+			testAccPreCheckFeatureSupported(t, features.MultipleApplicationSources)
+			testAccPreCheckFeatureSupported(t, features.ApplicationSourceName)
+		},
 		ProviderFactories: testAccProviders,
 		Steps: []resource.TestStep{
 			{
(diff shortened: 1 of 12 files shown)


jodhi:
The test for v2.13 failed, but kinda make sense since v2.13 not supporting named sources

argocd/resource_argocd_application_test.go
the-technat, line 1: Could you implement a feature gate similar to how it was done in https://github.com/argoproj-labs/terraform-provider-argocd/commit/e634913f8ba5cb18a120a4a043a9c7df5275a0f7? That way the CI test for `2.13` won't fail but skip and users get a proper error message if they want to use this feature on `2.13` or lower.

argocd/resource_argocd_application_test.go
```diff
@@ -2446,9 +2448,10 @@ resource "argocd_application" "helm_values_external" {
     }
 
     source {
+      name            = "wodpress-values"
```
the-technat, line 2451: Could you also add an AttributeCheck to the test checking if the field arrived in the TF state? See [here](https://github.com/argoproj-labs/terraform-provider-argocd/blob/main/argocd/resource_argocd_application_test.go#L1080-L1083) for an example.

the-technat (changes requested):
Hi @jodhi 

Thanks for your contribution! We gladly welcome it.

I left some comments of things that should be added. Would you mind taking a look at these?

argocd/resource_argocd_application_test.go
jodhi: Hi sure thing, will take a look. thanks for the reference

argocd/resource_argocd_application_test.go
jodhi: added the few attributeCheck that case some cases:
- all sources name present
- only one source has name

argocd/resource_argocd_application_test.go
jodhi: added the feature gate. it still return error, but with better message. is this expected?
Screenshot 2025-05-19 at 17 32 21

jodhi:
updated the PR, please help to review @the-technat . thank you

jodhi:
looks like the test failed, let me check again.
Just realized this part `testAccPreCheckFeatureSupported`, I will check this

jodhi:
added additional feature element on PreCheck and the test looks good on my local. thank you

mkilchhofer (approved):

the-technat (approved):

the-technat:
I review it again with my colleague and now it looks good.

Thanks again for your contribution @jodhi!

PR #659, opened 2025-05-28
Fix/helm empty blocks

the-technat:
After testing my fix in  https://github.com/argoproj-labs/terraform-provider-argocd/pull/656 I realized it doesn't catch the panic since `v` is a bit more complex than simply `nil`.

This PR reverts the fix and does it slightly different by silently returning an empty spec when the type-casting causes an error, thus fixing the bug.


mkilchhofer (approved):
LGTM. Lets hope we can fix this panic this time :D

PR #675, opened 2025-06-30
docs: render schema changes

blakepettersson:
#648 changed how docs were rendered. It seems like the schema was inadvertently frozen to the time when it was created. We don't need the templates for most cases, and for the parts we do we should dynamically render the schema instead.


mkilchhofer (approved):
LGTM 👍

templates/resources/account_token.md.tmpl
```diff
@@ -20,20 +20,6 @@ Manages ArgoCD [account](https://argo-cd.readthedocs.io/en/latest/user-guide/com
```
mkilchhofer, line 20: Is it intended that you want to keep this file here, @blakepettersson?

the-technat (approved):
Thanks for fixing this @blakepettersson! I was aware that it's currently somewhat static and broken, but hadn't looked into fixing it yet.

PR #680, opened 2025-07-11
chore: add possibility to use testcontainers

blakepettersson:
As a convenience, it's nice to be able to run acceptance tests directly without needing to setup the test environment first. While it's not _too_ burdensome, I'm kinda lazy and prefer having a single step to do this, and it's equally nice to not to have to pollute my own environment with yet another k8s cluster configuration.

This change is fully compatible with the existing offering and needs to be opted into by setting `USE_TESTCONTAINERS=true` when running `go test` (or by using the `make testacc_testcontainers` task).

As a part of this, I needed to rework the kustomizations slightly so that it's slightly more static (we now have overlays based on Argo CD versions), instead of relying on needing to pull the Argo CD manifests using curl.


internal/testhelpers/suite.go
```diff
@@ -0,0 +1,77 @@
...
+	// Setup the test environment once
+	var setupErr error
+
+	testEnvOnce.Do(func() {
+		argoCDVersion := os.Getenv("ARGOCD_VERSION")
+		if argoCDVersion == "" {
+			argoCDVersion = "v3.0.0"
+		}
+
+		k8sVersion := "v1.31.6"
```
the-technat, line 50: Maybe we want this to be configurable by the env var in the Makefile?

internal/testhelpers/suite.go
```diff
@@ -0,0 +1,77 @@
...
+	ctx, cancel := context.WithTimeout(context.Background(), DefaultTestTimeout)
+	defer cancel()
+
+	// Setup the test environment once
+	var setupErr error
+
+	testEnvOnce.Do(func() {
+		argoCDVersion := os.Getenv("ARGOCD_VERSION")
+		if argoCDVersion == "" {
+			argoCDVersion = "v3.0.0"
```
the-technat, line 47: Same as below?

blakepettersson:
@the-technat thanks for the review! I DRY-ed these settings so that they're now only taken from the `Makefile`. I also needed to tweak some of these tests so that they work when using testcontainers as well

Issue #613, opened 2025-04-03
`argocd_repository`


the-technat:

the-technat:
Has been implemented by @blakepettersson  in https://github.com/argoproj-labs/terraform-provider-argocd/pull/685

PR #683, opened 2025-07-14
feat: account resource/data source

blakepettersson:
This commit adds the possibility to create an account, as well as retrieving an account as a data source.


mkilchhofer:
Hi @blakepettersson 

Would it be possible that you implement the new resource and data source directly with the new approach using HashiCorp's plugin framework instead of SDKv2?

Otherwise even more human resources are required to migrate the code for all existing resources:
- https://github.com/argoproj-labs/terraform-provider-argocd/issues/609

You can find already 1 data source and 1 resource using the new approach:
- https://github.com/argoproj-labs/terraform-provider-argocd/blob/main/internal/provider/data_source_application.go
- https://github.com/argoproj-labs/terraform-provider-argocd/blob/main/internal/provider/resource_gpg_key.go

It's not a blocker for me as we still were unsuccessful with the migration of:
- #613

mkilchhofer:
@blakepettersson did you close that intentionally?

PR #698, opened 2025-08-01
chore: preload argocd before running testcontainer tests

blakepettersson:
While running this locally I got a few tests that were persistently failing. It turned out to be because pulling the argocd image was slow, timing out the test init. Now we ensure that the image is pulled before performing `kubectl wait`, so that we only need to wait for the pods to be ready.


the-technat:
Makes sense to me

PR #706, opened 2025-08-11
chore: set defaults in testmain

blakepettersson:
This keeps the default configuration in one place, and at the same time allows for acceptance tests to not to have to use the make targets to get the default values set (useful in e.g the IDE).

This also flips `USE_TESTCONTAINERS` to have a default of `true` (for now we don't do that in CI). This is something which can be discussed to see if that is desired by others.


blakepettersson:
Thank you @the-technat! 🙏

PR #712, opened 2025-08-21
fix: prevent deadlock

### LINKED ISSUE #710: argocd_repository resource stuck in endless update loop after upgrading to argoproj-labs/argocd ~> 7.0
## DESCRIPTION 
After migrating from the `oboukili/argocd` provider to` argoproj-labs/argocd` and updating the provider configuration, we encounter persistent issues with the `argocd_repository` resource.

Every terraform plan or apply shows unexpected changes for argocd_repository, even though no actual configuration changes were made. This results in an endless update loop where Terraform continuously tries to modify the resource but never completes successfully.

## Terraform Configuration

Provider 
```
terraform {
  required_providers {
    argocd = {
      version = "~> 7.0"
    }
  }
}
```
Resource (unchanged since migration)

```
resource "argocd_repository" "private" {
  count    = var.argocd_gitlab_token != "" ? 1 : 0
  name     = "gitlab-private"
  repo     = "https://gitlab.com/${var.argocd_gitlab_username}/123.git"
  username = var.argocd_gitlab_username
  password = var.argocd_gitlab_token
}
```

## Expected Behavior
Terraform should detect no changes for `argocd_repository` when configuration remains unchanged. 

## Actual Behavior

TERRAFORM INIT
```
> /opt/tofuenv/bin/tofu init

Initializing the backend...
Initializing modules...

Initializing provider plugins...
- terraform.io/builtin/terraform is built in to OpenTofu
- Reusing previous version of hashicorp/aws from the dependency lock file
- ------
- Reusing previous version of argoproj-labs/argocd from the dependency lock file
- Reusing previous version of hashicorp/kubernetes from the dependency lock file
- Using previously-installed hashicorp/kubernetes v2.38.0
- Using previously-installed statsig-io/statsig v2.1.1
- Using previously-installed hashicorp/random v3.7.2
- Using previously-installed gitlabhq/gitlab v17.11.0
- Using previously-installed hashicorp/aws v5.100.0
- Using previously-installed confluentinc/confluent v2.37.0
- Using previously-installed argoproj-labs/argocd v7.10.1   <-----
- Using previously-installed datadog/datadog v3.71.0

OpenTofu has been successfully initialized!
> /opt/tofuenv/bin/tofu validate
Success! The configuration is valid.

```
TERRAFORM PLAN
Terraform continuously attempts an in-place update for the resource:
```
  # module.deploy.argocd_repository.private[0] will be updated in-place
  ~ resource "argocd_repository" "private" {
      ~ connection_state_status = "Successful" -> (known after apply)
      ~ id                      = "https://gitlab.com/-----/123.git" -> (known after apply)
      ~ inherited_creds         = false -> (known after apply)
        name                    = "gitlab-private"
        # (7 unchanged attributes hidden)
    }
```

TERRAFORM APPLY
Terraform never finishes applying, resulting in:

`module.deploy.argocd_repository.private[0]: Still modifying... [id=-----, 1h18m30s elapsed]`

The only way forward is to abort

## Temporary Workaround

Downgrading the provider to v7.8.0 resolves the issue. With this version, Terraform no longer detects unnecessary changes and the apply succeeds. The workaround was found on this [issue](https://github.com/argoproj-labs/terraform-provider-argocd/issues/697)

However, this is only a temporary workaround. Ideally, we should not have to pin to a specific version but be able to safely use ~> 7.0.

## Notes
- The issue appears only on versions on version 7.10.1 for us 
- Pinning to ~> 7.0 is currently unsafe because of this behaviour.
- Ideally, the provider should support the latest versions without causing perpetual update loops.

blakepettersson:
Fixes #710. While migrating from an older version using sdkv2 to terraform-plugin-framework may need to recalculate the computed attributes, it should definitely not deadlock in the meantime.

Added a test where we create a repository using `7.8.0`, and then update it using the current version of this provider. Whereas before this deadlocked, it now updates as intended, and once this initial migration update has been done there is no state diff afterwards.

### DIFF
--- a/internal/provider/resource_repository.go
+++ b/internal/provider/resource_repository.go
@@ -195,10 +195,12 @@ func (r *repositoryResource) Update(ctx context.Context, req resource.UpdateRequ
 	}
 
 	var updatedRepo *v1alpha1.Repository
+
 	func() {
 		// Keep mutex enclosed in a function to keep the lock scoped to it and to prevent deadlocking
 		sync.RepositoryMutex.Lock()
 		defer sync.RepositoryMutex.Unlock()
+
 		updatedRepo, err = r.si.RepositoryClient.UpdateRepository(
 			ctx,
 			&repository.RepoUpdateRequest{Repo: repo},
--- a/internal/provider/resource_repository.go
+++ b/internal/provider/resource_repository.go
@@ -194,14 +194,16 @@ func (r *repositoryResource) Update(ctx context.Context, req resource.UpdateRequ
 		return
 	}
 
-	// Update repository
-	sync.RepositoryMutex.Lock()
-	defer sync.RepositoryMutex.Unlock()
-
-	updatedRepo, err := r.si.RepositoryClient.UpdateRepository(
-		ctx,
-		&repository.RepoUpdateRequest{Repo: repo},
-	)
+	var updatedRepo *v1alpha1.Repository
+	func() {
+		// Keep mutex enclosed in a function to keep the lock scoped to it and to prevent deadlocking
+		sync.RepositoryMutex.Lock()
+		defer sync.RepositoryMutex.Unlock()
+		updatedRepo, err = r.si.RepositoryClient.UpdateRepository(
+			ctx,
+			&repository.RepoUpdateRequest{Repo: repo},
+		)
+	}()
 
 	if err != nil {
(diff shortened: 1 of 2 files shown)


the-technat (approved):
Thanks @blakepettersson  for fixing this!

LGTM

PR #739, opened 2025-10-03
fix: replace bitnami helm repo with kubernetes-sigs one

the-technat:


argocd/resource_argocd_application_set_test.go
```diff
@@ -3322,9 +3322,9 @@ resource "argocd_application_set" "custom_namespace" {
       spec {
 		project = argocd_project.custom_namespace.metadata[0].name
 		source {
-		  repo_url        = "https://raw.githubusercontent.com/bitnami/charts/archive-full-index/bitnami"
-		  chart           = "redis"
-		  target_revision = "16.9.11"
+		  repo_url        = "https://kubernetes-sigs.github.io/descheduler"
```
the-technat, line 3325: Descheduler is a zero-dependency thing from kubernetes-sigs and hopefully around for a long time. It's deployed as a cronjob by default so deployment is fast.

argocd/resource_argocd_application_test.go
```diff
@@ -168,7 +168,7 @@ func TestAccArgoCDApplication_Helm_FileParameters(t *testing.T) {
 			{
 				Config: testAccArgoCDApplicationHelm_FileParameters(acctest.RandomWithPrefix("test-acc")),
 				// Setting up tests for this is non-trivial so it is easier to test for an expected failure (since file does not exist) than to test for success
-				ExpectError: regexp.MustCompile(`(?s)Error: failed parsing.*--set-file data`),
+				ExpectError: regexp.MustCompile(`(?s)Error:\s+failed\s+parsing.*--set-file data`),
```
the-technat, line 171: This ensures there could be multiple line spaces between the parts of the error message (as it's currently the case)

mkilchhofer (approved):

PR #747, opened 2025-10-08
fix(510): set server of argocd_cluster resource to required & force_new

the-technat:
**What type of PR is this?**

Fixes #?


the-technat:
A local test shows that it works as intended: 
image

mkilchhofer (approved):
LGTM

blakepettersson:
Could we have a test for this?

the-technat:
Sure, let me try to add one

onematchfox:
Just checking, is server actually required? Haven't gone to check the code but your screenshot shows id - which can be either server or name. 

Setting this to required now could be seen as a breaking change.

the-technat:
Good catch @onematchfox, it appears that `server` is the new default id, but it could also be set to the name:

image

That might have changed over the years...

I agree that this could be considered a breaking change. That's why I haven't merged yet but wanted some more feedback.

onematchfox:
No issue with setting `ForceNew: true` but maybe leave `Required` out?

the-technat:
Good point @onematchfox. I set it back to `Optional: false` to not break anything but added `ForceNew: true`.

@blakepettersson  I added a test that applies a resource, then reapplies it with another URL and expects the TF plan to require a replace of the resource.

onematchfox (approved):

PR #782, opened 2025-12-10
feat: Add ability to trigger sync on App (e.g. outside Sync window)

mkilchhofer:
**What type of PR is this?**

Fixes #?


the-technat:
image

Maybe set the field as omit-empty? Or ensure it's written to the state given that there's a default value.

PR #784, opened 2025-12-11
fix: set computed on namespace field

### LINKED ISSUE #783: Release 7.12.3 forces ArgoCD project replacement
The newly introduced release 7.12.3 forces project replacement due to metadata changes. Applying the update requires deleting all applications first which is not a desirable approach. Alternatively, appproject resources probably could be patched manually.

```
  # module.apps.argocd_project.tech must be replaced
-/+ resource "argocd_project" "tech" {
      ~ id = "tech" -> (known after apply)

      ~ metadata {
          ~ generation       = 1 -> (known after apply)
            name             = "tech"
          - namespace        = "argocd" -> null # forces replacement
          ~ resource_version = "15653" -> (known after apply)
          ~ uid              = "616f2bfa-b6ec-490f-82fd-4875ba23d1f0" -> (known after apply)
        }

        # (1 unchanged block hidden)
    }
```

```
Terraform version: OpenTofu v1.10.5
ArgoCD provider version: 7.12.3
ArgoCD version: 2.12.4
```

## Question
Was removal of namespace from metadata intentional? If so, what is recommended approach to handle this situation?

blakepettersson:
**What type of PR is this?**

/kind bug

If `namespace` is not set the API will return `argocd` as the namespace, since that's the default. To account for this we need to set the namespace field as computed, which is also what the previous implementation had.

We need to do something similar for `description`, where if we do not set it in TF, we should not set a `description` value.

Also took the opportunity to fix some incorrect docs, which if needed can be punted for another PR.

Fixes #783.

### DIFF
--- a/docs/data-sources/application.md
+++ b/docs/data-sources/application.md
@@ -39,7 +39,7 @@ data "argocd_application" "foo" {
 
 Required:
 
-- `name` (String) Name of the applications.argoproj.io, must be unique. Cannot be updated. More info: http://kubernetes.io/docs/user-guide/identifiers#names
+- `name` (String) Name of the applications.argoproj.io, must be unique. Cannot be updated. More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
 
 Optional:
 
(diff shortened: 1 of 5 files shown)


the-technat (approved):
Thanks @blakepettersson for looking into this!

mkilchhofer (approved):

PR #797, opened 2025-12-30
fix: add `type` for repocreds

### LINKED ISSUE #791: No type variable in resource argocd_repository_credentials

```
Terraform version: v1.14.1
ArgoCD provider version: v7.12.4
ArgoCD version: v3.1.5+cfeed49
```

### Affected Resource(s)
- argocd_repository_credentials

```hcl
resource "argocd_repository_credentials" "repo_creds" {
  url        = var.oci_repository
  username   = var.username
  password   = var.password
  # type = "helm" # Option Doesn't Exist
  enable_oci = true # OCI Fails because type is git by default
}
```

### Steps to Reproduce
1. `terraform apply` --> Creates `argocd_repository_credentials` without type `helm`
2. `terraform apply` --> Recreates `argocd_repository_credentials` because changes:
 `~ enable_oci = false -> true` forcing replace and producing the `inconsistent result after apply` error.

### Expected Behavior
Repository credentials should be able to be created with `type = "helm"` and `enable_oci = true` shouldn't recreate the `argocd_repository_credentials`.

This should work similar to the creation of `argocd_repository` with `type` argument.

It seems that the argocd provider creates the `argocd_repository_credentials` with `enable_oci`, however because `type` isn't `helm` it automatically detects when refreshing state that `enable_oci` is `false` and should be changed to `true`.

In-turn the provider tries to replace the object and it outputs the `inconsistent result after apply` error.

```
  # argocd_repository_credentials.repo_creds is tainted, so must be replaced
-/+ resource "argocd_repository_credentials" "repo_creds" {
      ~ enable_oci = false -> true
...

```
### Actual Behavior
```
argocd_repository_credentials.repo_creds: Creating...
│ Error: Provider produced inconsistent result after apply
│ 
│ When applying changes to argocd_repository_credentials.repo_creds, provider
│ "provider[\"registry.terraform.io/argoproj-labs/argocd\"]" produced an unexpected new value: .enable_oci: was cty.True,
│ but now cty.False.
│ 
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
```

### Important Factoids

Connection via `grpc_web = true`.

### Community Note

blakepettersson:
**What type of PR is this?**

/kind bug

Add `type` to repository_credentials. 

Apart from that we add a bit of a hack since Argo CD doesn't return
`type` nor `enable_oci` when listing repository credentials. Pending a real
fix in Argo CD, in the meantime we will just trust that any repository
credential that has been created/updated with a `type` has been properly 
saved, and we won't modify the state for those fields accordingly.

Fixes #791 (ish)

### DIFF
--- a/internal/provider/resource_repository_credentials.go
+++ b/internal/provider/resource_repository_credentials.go
@@ -208,6 +208,9 @@ func (r *repositoryCredentialsResource) Read(ctx context.Context, req resource.R
 	// Only overwrite if API explicitly returns true
 	if creds.EnableOCI {
 		result.EnableOCI = types.BoolValue(true)
+	} else if result.EnableOCI.IsNull() || result.EnableOCI.IsUnknown() {
+		// For import or initial read, set to default value if API returns false
+		result.EnableOCI = types.BoolValue(false)
 	}
 	// Otherwise keep the prior state value (API accepted it without error)
 


the-technat (approved):
Awesome fix @blakepettersson! Love the test you added and I'm absolutely okay "hacking" this in the provider as long as Argo CD itself behaves differently than we expect.

PR #812, opened 2026-01-28
Feature/use azure workload identity

vogtph:
addin workload identity to main


vogtph:
first to main of vogtph

PR #825, opened 2026-02-19
Add use_and_operation field to project resource sync_window block

ntrianta:
**What type of PR is this?**

/kind enhancement

Allows the `andOperator` field for project sync windows to be controlled via terraform

* [x] Documentation update is required by this PR.

Fixes #?

### DIFF
--- a/internal/provider/resource_project_test.go
+++ b/internal/provider/resource_project_test.go
@@ -393,6 +393,7 @@ resource "argocd_project" "simple" {
       duration = "3600s"
       schedule = "10 1 * * *"
       manual_sync = true
+	  use_and_operator = false
     }
     sync_window {
       kind = "deny"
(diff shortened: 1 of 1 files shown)


the-technat:
Thanks for your PR @ntrianta. Overall it looks almost good. You might want to split the test into a separate one so that it has less side-effects on other tests and you're likely missing the changes from `make generate`, that should update the TF docs for the project resource.

ntrianta:
> Thanks for your PR @ntrianta. Overall it looks almost good. You might want to split the test into a separate one so that it has less side-effects on other tests and you're likely missing the changes from `make generate`, that should update the TF docs for the project resource.

Hello! Thanks for the feedback @the-technat! I have fixed the documentation generation and the failing acceptance test. Regarding the splitting of the test, I can only see one one test specific to project sync windows, that checks fields that cannot error out (e.g. `namespace`, `cluster`, etc.), `TestAccArgoCDProject_SyncWindowsConsistency` . I don't mind adding a new test but it seems contrary to what is already there; please let me know what you prefer :)

ntrianta:
Thanks @the-technat!

PR #829, opened 2026-02-26
feat: enable proxy configuration for repositories

Tchebychev:
# feat(repository): add proxy and no_proxy configuration support

### Problem

When configuring repositories in ArgoCD, users may need to access Git or Helm repositories through a proxy server. According to the [ArgoCD documentation](https://argo-cd.readthedocs.io/en/stable/operator-manual/declarative-setup/#configure-repositories-with-proxy), the `proxy` and `noProxy` fields are supported in repository configuration.

However, the current Terraform provider does not expose these fields, preventing users from configuring proxy settings via Terraform.

### Solution

This PR adds support for `proxy` and `no_proxy` attributes to the `argocd_repository` resource, allowing users to configure HTTP/HTTPS proxy settings when accessing repositories.

### Changes

- **`internal/provider/model_repository.go`**:
  - Added `Proxy` and `NoProxy` fields to `repositoryModel` struct
  - Added schema attributes for `proxy` and `no_proxy` in `repositorySchemaAttributes()`
  - Updated `toAPIModel()` to map proxy fields to the ArgoCD API
  - Updated `updateFromAPI()` to read proxy values from API responses

- **`examples/resources/argocd_repository/resource.tf`**:
  - Added example usage with proxy configuration

- **`docs/resources/repository.md`**:
  - Updated documentation with new attributes and example

### Usage

```terraform
resource "argocd_repository" "with_proxy" {
  repo     = "https://github.com/example/repo.git"
  username = "git"
  password = "my-token"
  proxy    = "http://proxy.example.com:8080"
  no_proxy = "*.internal.example.com,localhost"
}
```

### New Attributes

| Attribute | Type | Description |
|-----------|------|-------------|
| `proxy` | String (Optional) | HTTP/HTTPS proxy to access the repository |
| `no_proxy` | String (Optional) | Comma-separated list of hostnames that should be excluded from proxying |

### Related Documentation

- [ArgoCD - Configure repositories with proxy](https://argo-cd.readthedocs.io/en/stable/operator-manual/declarative-setup/#configure-repositories-with-proxy)

### Checklist

- [x] Code follows the project's coding style
- [x] Documentation updated
- [x] Examples added
- [ ] Tests added/updated (if applicable)

### DIFF
--- a/internal/provider/resource_repository_test.go
+++ b/internal/provider/resource_repository_test.go
@@ -1139,53 +1139,23 @@ resource "argocd_repository" "global_to_project" {
 `, projectName, repoURL)
 }
 
-// TestAccArgoCDRepository_ProxyConsistency tests consistency of proxy and no_proxy fields
-// Note: This test uses a Helm repository which doesn't require a proxy but allows setting proxy fields
-func TestAccArgoCDRepository_ProxyConsistency(t *testing.T) {
-	config := `
-resource "argocd_repository" "proxy" {
-  repo     = "https://helm.nginx.com/stable"
-  name     = "nginx-stable-proxy-test"
-  type     = "helm"
-  proxy    = "http://proxy.example.com:8080"
-  no_proxy = "localhost,127.0.0.1"
-}
-`
-
+// TestAccArgoCDRepository_ProxyConnectivityError verifies that proxy configuration
+// is correctly passed to ArgoCD by expecting a connection failure when using an invalid proxy.
+func TestAccArgoCDRepository_ProxyConnectivityError(t *testing.T) {
 	resource.Test(t, resource.TestCase{
 		PreCheck:                 func() { testAccPreCheck(t) },
 		ProtoV6ProviderFactories: testAccProtoV6ProviderFactories,
 		Steps: []resource.TestStep{
 			{
-				Config: config,
-				Check: resource.ComposeAggregateTestCheckFunc(
-					resource.TestCheckResourceAttr(
-						"argocd_repository.proxy",
-						"proxy",
-						"http://proxy.example.com:8080",
-					),
-					resource.TestCheckResourceAttr(
-						"argocd_repository.proxy",
-						"no_proxy",
-						"localhost,127.0.0.1",
-					),
-				),
-			},
-			{
-				// Apply the same configuration again to test for consistency
-				Config: config,
-				Check: resource.ComposeAggregateTestCheckFunc(
-					resource.TestCheckResourceAttr(
-						"argocd_repository.proxy",
-						"proxy",
-						"http://proxy.example.com:8080",
-					),
-					resource.TestCheckResourceAttr(
-						"argocd_repository.proxy",
-						"no_proxy",
-						"localhost,127.0.0.1",
-					),
-				),
+				Config: `
+resource "argocd_repository" "proxy_fail" {
+  repo     = "https://helm.nginx.com/stable"
+  name     = "nginx-stable-proxy-fail"
+  type     = "helm"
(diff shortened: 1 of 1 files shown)


blakepettersson:
Hi, thanks for your contribution! Can you please provide:

* DCO check
* As well as acceptance tests making use of these fields

Tchebychev:
> Hi, thanks for your contribution! Can you please provide:
> 
> * DCO check
> * As well as acceptance tests making use of these fields

Hi, i think i'm done on those 2 requirements.

Tchebychev:
I've added a test TestAccArgoCDRepository_ProxyConsistency which configures a dummy proxy URL (http://proxy.example.com:8080) on a real public Helm repository.

Startups of acceptance tests seem to fail. I think that ArgoCD validates the connection upon creation/update. Since the proxy URL is fake/unreachable, the resource creation likely fails on the ArgoCD side, causing the test to fail.

Do you have a recommended way to test this safely in your CI environment?
- Should I use a specific reachable proxy URL available in your test environment?
- Or is there a way to force creation even if connection checks fail (similar to insecure or specific flags)?
- Or should I mock the validation differently ?

the-technat:
> I've added a test TestAccArgoCDRepository_ProxyConsistency which configures a dummy proxy URL (http://proxy.example.com:8080) on a real public Helm repository.
> 
> Startups of acceptance tests seem to fail. I think that ArgoCD validates the connection upon creation/update. Since the proxy URL is fake/unreachable, the resource creation likely fails on the ArgoCD side, causing the test to fail.
> 
> Do you have a recommended way to test this safely in your CI environment?
> 
> * Should I use a specific reachable proxy URL available in your test environment?
> * Or is there a way to force creation even if connection checks fail (similar to insecure or specific flags)?
> * Or should I mock the validation differently ?

The easiest way to solve this is to expect an expect an error in the apply of your resource, similar to [this one](https://github.com/argoproj-labs/terraform-provider-argocd/blob/main/argocd/resource_argocd_account_token_test.go#L134). 

Alternatively we could also deploy something like [mitmproxy](https://www.mitmproxy.org) as a container in the testing cluster and use this to connect, but that would be a bit more involved.

Tchebychev:
> > I've added a test TestAccArgoCDRepository_ProxyConsistency which configures a dummy proxy URL (http://proxy.example.com:8080) on a real public Helm repository.
> > Startups of acceptance tests seem to fail. I think that ArgoCD validates the connection upon creation/update. Since the proxy URL is fake/unreachable, the resource creation likely fails on the ArgoCD side, causing the test to fail.
> > Do you have a recommended way to test this safely in your CI environment?
> > 
> > * Should I use a specific reachable proxy URL available in your test environment?
> > * Or is there a way to force creation even if connection checks fail (similar to insecure or specific flags)?
> > * Or should I mock the validation differently ?
> 
> The easiest way to solve this is to expect an expect an error in the apply of your resource, similar to [this one](https://github.com/argoproj-labs/terraform-provider-argocd/blob/main/argocd/resource_argocd_account_token_test.go#L134).
> 
> Alternatively we could also deploy something like [mitmproxy](https://www.mitmproxy.org) as a container in the testing cluster and use this to connect, but that would be a bit more involved.

Let's try it with the ExpectError, i've done the changes

Tchebychev:
Thanks for your review :)

If everything looks good, it would be great to do a merge and a new release so I can start using this feature in my projects.
Do you think it's possible to make a release soon ? Maybe today ?

the-technat:
Sure, here we go: https://github.com/argoproj-labs/terraform-provider-argocd/actions/runs/22477920359/job/65108844483

PR #863, opened 2026-04-02
feat: add helm parameter_wo

the-technat:
**What type of PR is this?**

/kind feature

Adds write-only helm paramters. Useful if you set secrets this way that shouldn't be persisted into TF state

* [x] Documentation update is required by this PR.

Fixes #?


the-technat:
According to  addhttps://developer.hashicorp.com/terraform/plugin/sdkv2/resources/write-only-arguments#schema-behavior you can't set TypeSet blocks to write-only:
image

PR #867, opened 2026-04-14
feat(argocd_applicationset): add exclude field to Git Generator for file

HarshPanchal18:
**What type of PR is this?**

/kind enhancement

There was given a configuration variable `exclude` for Git Generators of directory but not for Git Generators of file. This PR enhances the Git generator to configure `exclude` from terraform.

Fixes #847

### DIFF
--- a/argocd/resource_argocd_application_set_test.go
+++ b/argocd/resource_argocd_application_set_test.go
@@ -151,12 +151,11 @@ func TestAccArgoCDApplicationSet_gitFiles(t *testing.T) {
 					),
 					resource.TestCheckResourceAttrSet(
 						"argocd_application_set.git_files",
-						"spec.0.generator.0.git.0.file.0.exclude",
+						"spec.0.generator.0.git.0.file.1.path",
 					),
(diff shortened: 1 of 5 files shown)


HarshPanchal18:
Can someone look into the failing test-cases issue as it doesn't seems problematic because of the changes I made. If it is, guide me pls.

the-technat:
Your PR contains ~49 commits from main, seems like there went something wrong with rebasing. Could you fix this first?

the-technat (commented):
Have you tested what happens to existing TF resources when switching to a provider version that contains this change? e.g does it work or will the next TF plan run with the new provider generate a diff that's based on defaulting this argument to `false` without me changing anything (which would be undesired)?

HarshPanchal18:
This `exclude` for file will work very similar to the `exclude` for the directory.

These changes won't cause unexpected changes to current resources managed by Terraform and will not generate any diff for TF plan until and unless the value(`exclude = true`) is not provided explicitly in `resource.tf`.

the-technat (approved):

PR #899, opened 2026-06-26
fix: prevent panic when removing roles from argocd_project on update

### LINKED ISSUE #898: Provider panics on `argocd_project` update when removing roles (index out of range)
```
Terraform version: 1.15.5
ArgoCD provider version: 7.15.3
ArgoCD version: 3.4.4
```

### Affected Resource(s)
- argocd_project

```hcl
# 1. Apply this first — project with 3 roles
resource "argocd_project" "this" {
  metadata {
    name      = "test-bug"
    namespace = "argocd"
  }
  spec {
    description  = "repro"
    source_repos = ["*"]
    destination {
      name      = "in-cluster"
      namespace = "*"
    }
    role {
      name     = "role-a"
      policies = ["p, proj:test-bug:role-a, applications, get, test-bug/*, allow"]
    }
    role {
      name     = "role-b"
      policies = ["p, proj:test-bug:role-b, applications, get, test-bug/*, allow"]
    }
    role {
      name     = "role-c"
      policies = ["p, proj:test-bug:role-c, applications, get, test-bug/*, allow"]
    }
  }
}

# 2. Then remove role-a and role-b, keeping only role-c, and apply again -> panic
#    (role-c sits at index 2 in the existing project, but at index 0 in the new spec)
```

### Panic Output
```
panic: runtime error: index out of range [5] with length 5

goroutine 137 [running]:
github.com/argoproj-labs/terraform-provider-argocd/internal/provider.(*projectResource).Update(...)
        github.com/argoproj-labs/terraform-provider-argocd/internal/provider/resource_project.go:384

Error: The terraform-provider-argocd_v7.15.3 plugin crashed!
```

### Steps to Reproduce
1. `terraform apply` the config above with 3 roles.
2. Edit the config to remove the leading role(s), keeping a role that had a higher index in the existing project (e.g. keep only `role-c`).
3. `terraform apply` again → the provider panics.

### Expected Behavior
The update succeeds and pre-existing JWT tokens for retained managed roles are preserved.

### Actual Behavior
The provider panics with `index out of range`.

**Root cause:** In `internal/provider/resource_project.go`, the JWT-preservation loop in `Update` uses the index `i` returned by `p.GetRoleByName(r.Name)` — which refers to the *existing* project's role slice — to index `spec.Roles`, the *new* desired spec. When roles are removed, the existing index can exceed `len(spec.Roles)`, causing the out-of-range panic. The fix is to index `spec.Roles` by the loop position in the desired spec rather than by the existing project's index.

### Important Factoids
Triggers whenever role count/order differs between the existing project and the new spec such that a retained role's old index is greater than or equal to the new role count.

### Community Note
* Please vote on this issue by adding a 👍 reaction to the original issue to help the community and maintainers prioritize this request

jabbrwcky:
**What type of PR is this?**

/kind bug

Fixes a panic (`runtime error: index out of range`) when updating an `argocd_project` that removes one or more `role` blocks. The JWT-preservation loop in `projectResource.Update` used the index returned by `p.GetRoleByName()` — which refers to the *existing* project's role slice — to index `spec.Roles` (the new desired spec). When roles are removed, that index can exceed `len(spec.Roles)`, crashing the provider. The fix indexes `spec.Roles` by the loop position in the desired spec instead.

(No docs change — behavior fix only.)

Fixes #898

Added acceptance test `TestAccArgoCDProjectRemoveLeadingRoles`: creates a project with 3 roles, then updates keeping only the last role (index 2 in the existing project, index 0 in the new spec). Panics on `main`, passes with the fix.

```
make testacc TESTARGS='-run TestAccArgoCDProjectRemoveLeadingRoles'
```

### DIFF
--- a/internal/provider/resource_project.go
+++ b/internal/provider/resource_project.go
@@ -389,7 +389,7 @@ func (r *projectResource) Update(ctx context.Context, req resource.UpdateRequest
 
 	// Preserve preexisting JWTs for managed roles
 	roles := expandProjectRoles(ctx, data.Spec[0].Role)
-	for _, r := range roles {
+	for specIdx, r := range roles {
 		var pr *v1alpha1.ProjectRole
 
 		var i int
@@ -406,8 +406,10 @@ func (r *projectResource) Update(ctx context.Context, req resource.UpdateRequest
 				return
 			}
 		} else {
-			// Only preserve preexisting JWTs for managed roles if we found an existing matching project
-			spec.Roles[i].JWTTokens = pr.JWTTokens
+			// Only preserve preexisting JWTs for managed roles if we found an existing matching role.
+			// Note: i indexes the existing project's roles (p), so we must write to spec.Roles using
+			// the desired-spec index (specIdx), which is aligned with the roles slice we are iterating.
+			spec.Roles[specIdx].JWTTokens = pr.JWTTokens
 		}
 	}
 
--- a/internal/provider/resource_project.go
+++ b/internal/provider/resource_project.go
@@ -387,9 +387,7 @@ func (r *projectResource) Update(ctx context.Context, req resource.UpdateRequest
 		return
 	}
 
-	// Preserve preexisting JWTs for managed roles
-	roles := expandProjectRoles(ctx, data.Spec[0].Role)
-	for specIdx, r := range roles {
+	for specIdx, r := range spec.Roles {
 		var pr *v1alpha1.ProjectRole
 
 		var i int
--- a/internal/provider/resource_project_test.go
+++ b/internal/provider/resource_project_test.go
@@ -180,6 +180,47 @@ func TestAccArgoCDProjectUpdateAddRole(t *testing.T) {
 	})
 }
 
+func TestAccArgoCDProjectRemoveLeadingRoles(t *testing.T) {
+	// Regression test: removing roles so that a retained role keeps a higher
+	// index in the existing project than its position in the new desired spec
+	// used to panic with "index out of range" while preserving JWT tokens.
+	name := acctest.RandomWithPrefix("test-acc")
+
+	resource.Test(t, resource.TestCase{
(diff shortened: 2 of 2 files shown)


internal/provider/resource_project.go
```diff
@@ -389,7 +389,7 @@ func (r *projectResource) Update(ctx context.Context, req resource.UpdateRequest
 
 	// Preserve preexisting JWTs for managed roles
 	roles := expandProjectRoles(ctx, data.Spec[0].Role)
-	for _, r := range roles {
+	for specIdx, r := range roles {
```
blakepettersson: I think we can just iterate `spec.Roles` directly 

```suggestion
	for specIdx, r := range spec.Roles {
```

jabbrwcky: I did not have time yet too look too deep into that but the role expansion produces a distinct slice.

The current code also assigns the found JWTs to `spec.Roles[...]`, so iterating directly over spec.Roles might lose the exiting JWTs: https://github.com/jabbrwcky/terraform-provider-argocd/blob/af9ff006b2f12b81197969ec5b52d35fa9723c49/internal/provider/resource_project.go#L412

The original bug was caused by nonexisting indices after all.

jabbrwcky: Ok, I took a look. 

The invocation of `expandProjectRoles` here is redundant here because it is already invoked by [`expandProject`](https://github.com/jabbrwcky/terraform-provider-argocd/blob/af9ff006b2f12b81197969ec5b52d35fa9723c49/internal/provider/resource_project.go#L351)  in line [692](https://github.com/jabbrwcky/terraform-provider-argocd/blob/af9ff006b2f12b81197969ec5b52d35fa9723c49/internal/provider/resource_project.go#L692)

Tests still pass so I think it is safe to change

jabbrwcky: Changed manually

blakepettersson (approved):
LGTM, thanks!

PR #900, opened 2026-06-26
feat(argocd_applicationset): add values field to pull_request generator

vtiruv:
**What type of PR is this?**

/kind enhancement

The `values` field is available on the `plugin`, `git`, `clusters` and `cluster_decision_resource` generators but not on `pull_request`, even though the upstream `PullRequestGenerator` type supports it. This PR adds `values` to the `pull_request` generator so it can be set from Terraform.

This is needed for the `merge(list, pull_request)` pattern, where `list` provides a baseline parameter set and `pull_request` overlays per-PR data on a shared mergekey. Without `values` on `pull_request` there is no way to set that key. 

N/A

Added a unit test for the expand/flatten round-trip and extended `TestAccArgoCDApplicationSet_pullRequestGithub` to set and check `values`. The field is optional and additive, so existing configurations are unaffected.

### DIFF
--- a/argocd/structure_application_set.go
+++ b/argocd/structure_application_set.go
@@ -522,6 +522,10 @@ func expandApplicationSetPullRequestGeneratorGenerator(mg interface{}, featureMu
 		asg.PullRequest.Template = temp
 	}
 
+	if v, ok := m["values"]; ok {
+		asg.PullRequest.Values = expandStringMap(v.(map[string]interface{}))
+	}
+
 	return asg, nil
 }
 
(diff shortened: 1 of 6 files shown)


argocd/structure_application_set.go
```diff
@@ -522,6 +522,10 @@ func expandApplicationSetPullRequestGeneratorGenerator(mg interface{}, featureMu
 		asg.PullRequest.Template = temp
 	}
 
+	if v, ok := m["values"]; ok {
```
the-technat: Is there a reason you're doing the cast in the call to `expandStringMap` and not within the if-clause as done by all the other checks? If not I'd suggest moving this up to not only check if the values key is present but also it it's a string-map.

the-technat (changes requested):

argocd/structure_application_set.go
vtiruv: Good catch, no real reason. I just copied the pattern from the other values blocks in the file (clusters/git/plugin), which all cast inside the call. However, I am happy to conform to match the local function style, will follow up with a commit.

the-technat (approved):
Thanks for your work!

PR #916, opened 2026-08-01
feat(provider): add server_name to override ArgoCD server pod selector for port forwarding

### LINKED ISSUE #555: Possible to change server name when using port forward?

```
Terraform version: 1.10.5
ArgoCD provider version: 7.3.1
ArgoCD version: 2.14.2
```

## Terraform configuration
```hcl
terraform {
  required_providers {
    argocd = {
      version = "7.3.1"
    }
  }
}

provider "argocd" {
  port_forward_with_namespace = "argocd"
  username                    = "admin"
  password                    = var.argocd_admin_password
}
```

## Question
I have installed ArgoCD with the bitnami helm chart so the server name is `argocd-argo-cd-server` so when trying to use the above provider config i get the error
```
cannot find pod with selector: [app.kubernetes.io/name=argocd-server] - use
the --{component}-name flag in this command or set the environmental
variable (Refer to
https://argo-cd.readthedocs.io/en/stable/user-guide/environment-variables),
to change the Argo CD component name in the CLI
```
This is being run in a CI pipeline on an unexposed api server. I have tried to set the `ARGOCD_SERVER_NAME` environment variable but I still get the above error. Port forwarding using `argocd-argo-cd-server` locally works. Is there anyway to set the server name from this provider?

pujitha24:
/kind enhancement

Motivation:
When using port_forward or port_forward_with_namespace, the provider
locates the ArgoCD API server pod using a hardcoded label selector value
of "argocd-server". Installations that rename this component (e.g. the
Bitnami Helm chart, which names it argocd-argo-cd-server) fail with
"cannot find pod with selector: [app.kubernetes.io/name=argocd-server]"
and have no way to work around it, since the value was never threaded
through to the underlying apiclient.ClientOptions.ServerName field that
the upstream argocd CLI already exposes via --server-name /
ARGOCD_SERVER_NAME (argoproj-labs/terraform-provider-argocd#555).

Approach:
Add an optional server_name provider attribute (mirrored in both the
SDKv2 schema in argocd/provider.go and the terraform-plugin-framework
schema in internal/provider/provider.go, which must stay in sync since
both are combined via tf6muxserver). setPortForwardingOpts in both
model_provider.go files now resolves ServerName from the explicit
attribute, then the ARGOCD_SERVER_NAME environment variable (matching
the upstream CLI's own env var), falling back to the previous
"argocd-server" default when neither is set. Existing configurations
are unaffected.

Validation:
- go build ./...
- USE_TESTCONTAINERS=false go test ./... (unit tests; this sandbox has
  no container runtime, so the Docker-based K3s+ArgoCD acceptance suite
  used by `make testacc` could not be run here)
- Added argocd/model_provider_test.go and
  internal/provider/model_provider_test.go, unit-testing
  setPortForwardingOpts's ServerName resolution for: unset (defaults to
  "argocd-server"), explicit provider config, and the ARGOCD_SERVER_NAME
  env var fallback. Confirmed these tests fail against the prior
  hardcoded implementation and pass against this change.
- Manually verified (via a throwaway test invoking the muxed provider's
  GetProviderSchema) that the SDKv2 and framework schemas remain
  compatible with the new attribute present in both.

Documentation has been updated (docs/index.md, examples/provider/provider.tf).

Fixes #555

### DIFF
--- a/argocd/model_provider.go
+++ b/argocd/model_provider.go
@@ -31,6 +31,7 @@ type ArgoCDProviderConfig struct {
 	ServerAddr               types.String `tfsdk:"server_addr"`
 	PortForward              types.Bool   `tfsdk:"port_forward"`
 	PortForwardWithNamespace types.String `tfsdk:"port_forward_with_namespace"`
+	ServerName               types.String `tfsdk:"server_name"`
 	Kubernetes               []Kubernetes `tfsdk:"kubernetes"`
 
 	// Run ArgoCD API server locally
@@ -235,7 +236,11 @@ func (p ArgoCDProviderConfig) setPortForwardingOpts(ctx context.Context, opts *a
 		}
 
 		opts.ServerAddr = "localhost" // will be overwritten by ArgoCD module when we initialize the API client but needs to be set here to ensure we
-		opts.ServerName = "argocd-server"
+
+		opts.ServerName = getDefaultString(p.ServerName, "ARGOCD_SERVER_NAME")
+		if opts.ServerName == "" {
+			opts.ServerName = "argocd-server"
+		}
 
 		if opts.PortForwardNamespace == "" {
 			opts.PortForwardNamespace = "argocd"
--- a/argocd/model_provider_test.go
+++ b/argocd/model_provider_test.go
@@ -1,0 +1,63 @@
+package argocd
+
+import (
+	"context"
+	"testing"
+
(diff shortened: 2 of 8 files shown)


pujitha24:
Just checking in on this one — it's still green and rebased, happy to make any changes that would help it move forward.

the-technat (approved):
Thanks for your contribution! Looks good to me.

PR #707, opened 2025-08-13
chore: migrate project and project tokens

blakepettersson:
Migrate project and project token resources to the terraform-plugin-framework.

Closes #776 
Closes #777

PR #920, opened 2026-08-05
fix(project): restore role policy validation and align resources with Argo CD

### LINKED ISSUE #345: need to add other resources in validResources of rbac
### Description

Need to include other resources like `applicationsets, extension, certificates ..` in the validResources for the rbac

### Potential Terraform Configuration

So far it seems only the below resources are supported through this provider

```hcl
var validResources = map[string]bool{
	"applications": true,
	"repositories": true,
	"clusters":     true,
	"exec":         true,
	"logs":         true,
}
```

### References
 

I don't see any issues wrt to this.

### Community Note

somaz94:
**What type of PR is this?**

/kind bug

`argocd_project` stopped validating role policies when the resource was migrated to the plugin framework in #707. The SDKv2 code called `validatePolicy` from `argocd/structure_project.go` while expanding the project, but the framework `expandProjectRoles` copies policies straight through. The helpers left behind in `argocd/utils.go` (`validatePolicy`, `isValidPolicyAction`, `validResources`) have had no production caller since, only a unit test.

This was easy to miss because the existing acceptance step still passes: Argo CD's own `AppProject.ValidateProject` rejects the same policy with the same "invalid policy rule" wording, so the failure just moved from plan time to apply time.

This PR restores the check as a `ConfigValidator` on the project resource, so an invalid policy is reported during `terraform validate` / `plan` instead of after a round trip to the API.

While restoring it, the accepted resource set is now derived from `rbac.ProjectScoped` instead of being re-listed in the provider. That is the set Argo CD actually enforces for project role policies, and it resolves #345:

* `applications`, `applicationsets`, `repositories`, `exec`, `logs` and `clusters` are accepted, which matches the previous behaviour in practice.
* `projects` is no longer accepted. It is a valid resource in the global RBAC config map but it is not project scoped, so Argo CD rejects it inside a role policy. The provider used to be more permissive than the server here.
* The resources named in #345 (`certificates`, `extensions`, `accounts`, `gpgkeys`, `write-repositories`) are still rejected, for the same reason. They are only meaningful in the global RBAC config map.

#345 also suggested using code exported by Argo CD where possible. `rbac.ProjectScoped` is exported, so the resource set can no longer drift. The action list still has to be duplicated here because Argo CD does not export the one it uses for project policies.

`make generate` produces no diff. The change adds a config validator and does not alter the schema.

Fixes #345

```
make build             # ok
make fmt               # no diff
make generate          # no diff
make lint              # no new findings
make testacc TEST_FILTER=TestAccArgoCDProject
                       # ok, 31 passed / 1 skipped (version gated), coverage 41.0%
```

Run against Argo CD v3.4.6 via the testcontainers K3s environment.

Two test layers were added:

* `internal/validators/project_policy_test.go` covers the policy rules themselves, including a case per entry of `rbac.ProjectScoped` so the accepted set cannot silently drift from Argo CD's.
* `internal/provider/resource_project_config_validation_test.go` runs the resource's config validators against a synthetic configuration. This is deliberately separate from the acceptance tests, which cannot tell a plan time rejection apart from an apply time one because the server produces the same message.

The unit test for the old helpers moved out of `argocd/utils_test.go` along with the helpers, since nothing calls them any more.

`resource_project_test.go` needed one adjustment. Five of the error-expecting configs declare a role named `incorrect role name` or `incorrect-syncwindow` but give it a policy whose subject is `proj:<project>:testrole`. Those configs each mean to isolate a single defect, and the mismatched subject only went unnoticed because the policy was not being checked. With the check restored, the policy error fires before the defect the step is actually asserting on, so the subjects are now aligned with their role names. No assertion was weakened.

On the behaviour change: every policy literal already present in this repo's acceptance tests, examples and docs uses only project scoped resources, so no existing configuration is affected. A user who had a `projects` policy in a role would previously have had it rejected by the API on apply, and will now see it at plan time instead.

One thing I noticed while working on this, not addressed here: `RoleNameValidator` and `SyncWindowDurationValidator` exist in `internal/validators/project_validators.go` but are not attached to the schema, so they are dead in the same way the policy check was. The related acceptance steps still pass because Argo CD rejects those values on apply. Happy to follow up in a separate PR if you would like them wired up. `RoleNameValidator` would also need its message reconciled with the lowercase wording the acceptance step matches on.


internal/validators/project_policy.go
```diff
@@ -0,0 +1,221 @@
...
+
+var _ resource.ConfigValidator = projectPolicyValidator{}
+
+// policyComponentCount is the number of comma separated components in a casbin
+// policy rule: 'p, sub, res, act, obj, eft'.
+const policyComponentCount = 6
+
+// validPolicyActions and validPolicyActionPatterns mirror the equivalent lists
+// in Argo CD's AppProject validation. Argo CD does not export them, so they
+// have to be kept in sync by hand.
```
the-technat, line 24: Maybe also add a link to the Argo CD docs/reference

the-technat (approved):
Thanks for putting your time into this! One more reason to finally finish the migration to the plugin framework, every resource that get's migrated shows new surprises 😕 and makes things more complex than it needs to be. The code looks good to me. Given that Argo CD already validates the same thing server-side I don't expect any breaking changes from this one, just better UX.

blakepettersson (approved):

PR #925, opened 2026-08-13
fix: give spec.sources precedence over spec.source when flattening

### LINKED ISSUE #909: `terraform import` of `argocd_application` imports `spec.source` and excludes `spec.sources`

```text
Terraform version: v1.15.8
ArgoCD provider version: v7.15.3
ArgoCD version: v2.14.11+8283115 and v3.4.1 (reproduced issue in both versions)
```

### Affected Resource(s)

- `argocd_application`

```hcl
terraform {
  required_providers {
    argocd = {
      version = "~> 7.0"
    }
  }
}

provider "argocd" {
  server_addr = "my-argocd-server.example.com"
}

resource "argocd_application" "example" {
  metadata {
    name      = "example-app"
    namespace = "argocd"
  }

  spec {
    project = "default"

    source {
      name            = "single"
      repo_url        = "https://github.com/bogus/single-source.git"
      path            = "single"
      target_revision = "main"
    }

    source {
      name            = "multi-source-a"
      repo_url        = "https://github.com/bogus/multi-source-a.git"
      path            = "multi-source-a"
      target_revision = "main"
    }

    source {
      name            = "multi-source-b"
      repo_url        = "https://github.com/bogus/multi-source-b.git"
      path            = "multi-source-b"
      target_revision = "main"
    }

    destination {
      server    = "https://kubernetes.default.svc"
      namespace = "default"
    }
  }
}
```

### Debug Output

N/A

### Panic Output

N/A

### Steps to Reproduce

- Create the ArgoCD `Application` via K8s:

  ```bash
  kubectl apply -f - <<'EOF'
  apiVersion: argoproj.io/v1alpha1
  kind: Application
  metadata:
    name: example-app
    namespace: argocd
  spec:
    project: default
    destination:
      server: https://kubernetes.default.svc
      namespace: default
    source:
      name: single
      repoURL: https://github.com/bogus/single-source.git
      path: single
      targetRevision: main
    sources:
      - name: multi-source-a
        repoURL: https://github.com/bogus/multi-source-a.git
        path: multi-source-a
        targetRevision: main
      - name: multi-source-b
        repoURL: https://github.com/bogus/multi-source-b.git
        path: multi-source-b
        targetRevision: main
  EOF
  ```

- Run `terraform import argocd_application.example example-app:argocd`

- Run `terraform plan`

### Expected Behavior

Preferred - the imported state logic should align to the ArgoCD logic and ignore the singular `spec.source` block. As per [ArgoCD docs](https://github.com/argoproj/argo-cd/blob/master/docs/user-guide/multiple_sources.md):

> You can provide multiple sources using the sources field. When you specify the sources field, ArgoCD will ignore the source (singular) field.

Alternative - the `terraform import` logic should import the full application spec from the K8s yaml, including `spec.source` and `spec.sources` into `source` blocks.

Either behavior would be an improvement over the current state; happy to follow maintainer preference.

### Actual Behavior

The imported state only contains the singular `spec.source` block from the K8s yaml. It does not include the `spec.sources` block.

`terraform state show argocd_application.example` snippet:

```
resource "argocd_application" "example" {
    id     = "example-app:argocd"

    <SNIP>

    spec {
        project                = "default"
        revision_history_limit = 0

        destination {
            name      = null
            namespace = "default"
            server    = "https://kubernetes.default.svc"
        }

        source {
            chart           = null
            name            = "single"
            path            = "single"
            ref             = null
            repo_url        = "https://github.com/bogus/single-source.git"
            target_revision = "main"
        }
    }
}
```

`terraform plan` proposes re-adding the missing source blocks:

```bash
argocd_application.example: Refreshing state... [id=example-app:argocd]

Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following
symbols:
  ~ update in-place

Terraform will perform the following actions:

  # argocd_application.example will be updated in-place
  ~ resource "argocd_application" "example" {
      + cascade  = true
        id       = "example-app:argocd"
      + validate = true
      + wait     = false
        # (1 unchanged attribute hidden)

      ~ spec {
          ~ revision_history_limit = 0 -> 10
            # (1 unchanged attribute hidden)

          + source {
              + name            = "multi-source-a"
              + path            = "multi-source-a"
              + repo_url        = "https://github.com/bogus/multi-source-a.git"
              + target_revision = "main"
            }
          + source {
              + name            = "multi-source-b"
              + path            = "multi-source-b"
              + repo_url        = "https://github.com/bogus/multi-source-b.git"
              + target_revision = "main"
            }

            # (2 unchanged blocks hidden)
        }

        # (1 unchanged block hidden)
    }

Plan: 0 to add, 1 to change, 0 to destroy.
```

### Important Factoids

As per [ArgoCD docs](https://github.com/argoproj/argo-cd/blob/master/docs/user-guide/multiple_sources.md):

> You can provide multiple sources using the sources field. When you specify the sources field, ArgoCD will ignore the source (singular) field.

That means the effective application behaviour should be driven by `spec.sources` when both are present.

In `resource_argocd_application.go`, import is configured with `schema.ImportStatePassthroughContext`, and the normal resource read then calls `flattenApplication` to populate state.

In `structure_application.go`, `flattenApplicationSpec` does:

```go
if s.Source != nil {
    spec["source"] = flattenApplicationSource([]application.ApplicationSource{*s.Source})
} else {
    spec["source"] = flattenApplicationSource(s.Sources)
}
```

which effectively makes `spec.source` win over `spec.sources` during flattening.

That does not match the ArgoCD docs expectation above, and also matches what I am seeing after import: only the singular `source` block is kept in Terraform state.

Noting this code looks shared with `terraform plan/apply` logic.

### References

- [resource_argocd_application.go](https://github.com/argoproj-labs/terraform-provider-argocd/blob/main/argocd/resource_argocd_application.go)
- [structure_application.go](https://github.com/argoproj-labs/terraform-provider-argocd/blob/main/argocd/structure_application.go)
- [ArgoCD multiple sources docs](https://github.com/argoproj/argo-cd/blob/master/docs/user-guide/multiple_sources.md)

### Community Note

- Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
- If you are interested in working on this issue or have submitted a pull request, please leave a comment

pujitha24:
Motivation:
When an ArgoCD Application object has both spec.source (singular) and
spec.sources (list) populated, flattenApplicationSpec always preferred
the singular source when writing Terraform state. This does not match
ArgoCD's own behavior: (*ApplicationSpec).HasMultipleSources() in
argo-cd/v3 gives Sources priority whenever len(Sources) > 0, and both
GetSource() and GetSources() effectively ignore Source in that case.

The mismatch is externally visible after `terraform import`: importing
an Application that has both fields set (as can happen with objects
created/edited outside Terraform) drops the multi-source configuration
from state entirely, and the next `terraform plan` proposes re-adding
the "missing" source blocks even though nothing changed in ArgoCD.

Approach:
In argocd/structure_application.go, flattenApplicationSpec now checks
len(s.Sources) > 0 first and flattens Sources in that case, falling
back to the singular Source only when Sources is empty. This mirrors
ArgoCD's own precedence exactly.

The write path (expandApplicationSpec / resource_argocd_application.go)
already normalizes user config into a single Sources slice (using
Source only when there's exactly one entry), so Terraform-managed
resources never produce an object with both fields populated - only
imported or externally-mutated objects can, which is exactly the
scenario this fixes.

Validation:
- go build ./... succeeds.
- USE_TESTCONTAINERS=false go test -v -cover ./argocd/... passes,
  including two new targeted tests in the new
  argocd/structure_application_test.go:
  TestFlattenApplicationSpecSourcesPrecedenceOverSource and
  TestFlattenApplicationSpecFallsBackToSingularSource. The first
  reproduces the exact bug: with both Source and Sources set, it
  fails before this change (flattened to 1 source, the singular one)
  and passes after (flattened to the 2 sources).
- golangci-lint run ./argocd/... shows only pre-existing goconst
  warnings in unrelated code, none on changed lines.
- gofmt -l reports no issues on the changed/added files.
- Did not run `make testacc` (requires a live K3s/ArgoCD cluster via
  testcontainers/kind, unavailable in this environment). This change
  is a pure flatten-logic fix fully exercised by the unit tests above,
  which reproduce the reported symptom directly.

Report: https://github.com/argoproj-labs/terraform-provider-argocd/issues/909

Fixes #909


argocd/structure_application_test.go
```diff
@@ -0,0 +1,74 @@
+package argocd
+
+import (
+	"testing"
+
+	application "github.com/argoproj/argo-cd/v3/pkg/apis/application/v1alpha1"
+)
+
+func TestFlattenApplicationSpecSourcesPrecedenceOverSource(t *testing.T) {
```
onematchfox: Please use Terraform Acceptance Tests rather than unit tests as per the standard in this repo 🙏

argocd/structure_application.go
```diff
@@ -633,10 +633,15 @@ func flattenApplicationSpec(s application.ApplicationSpec) []map[string]interfac
...
+	// ArgoCD ignores the singular `source` field whenever `sources` is set, see
+	// (*ApplicationSpec).HasMultipleSources() upstream. Mirror that precedence here so
+	// state reflects the sources ArgoCD actually syncs, rather than always preferring Source.
+	if len(s.Sources) > 0 {
+		spec["source"] = flattenApplicationSource(s.Sources)
+	} else if s.Source != nil {
 		spec["source"] = flattenApplicationSource([]application.ApplicationSource{*s.Source})
 	} else {
-		spec["source"] = flattenApplicationSource(s.Sources)
+		spec["source"] = flattenApplicationSource(nil)
```
onematchfox, line 644: This shouldn't be possible - an application should always have at least 1 source or sources shouldn't it? Is there validation missing somewhere else in the provider that would allow this situation to occur?

onematchfox (changes requested):

argocd/structure_application.go
the-technat: image
the Argo CD Swagger UI doesn't mark either of them required. Also as @pujitha24 already pointed out our schema requires one source to be present. In the rare case that someone imports an app that has no sources the provider should still be able to cope with it, as by this else case (haven't check other places we deal with sources though). Whether such an app can be imported or not depends on whether Argo CD accepts an app without sources, but that's irrelevant for us I think, so fine for me.

the-technat (approved):

onematchfox (approved):
Sorry - was happy that the conversation had been had/resolved. Thought @the-technat's approval was all that was needed - didn't realise I was still blocking this.

PR #926, opened 2026-08-13
fix(application): avoid nil pointer dereference reading sync policy automated fields

### LINKED ISSUE #919: Plugin did not respond / crash reading data.argocd_application right after apply on v7.16.0
Hi,

thanks for maintaining this provider! I ran into a crash and wanted to report it.

OpenTofu version: <bitte aus Spacelift Run-Log ablesen>
ArgoCD provider version: 7.16.0 (issue disappears when pinned to 7.15.3)
ArgoCD version: v3.4.5

### Affected Resource(s)

data.argocd_application

resource "null_resource" "argocd_sync" {
  triggers = {
    apps            = join(" ", var.apps)
    apps_reversed   = join(" ", reverse(var.apps))
    argocd_server   = var.argocd_server
    argocd_user     = "terraform"
    argocd_password = var.argocd_password
    always_run      = timestamp()
  }

  provisioner "local-exec" {
    command = templatefile("${path.module}/sync.sh.tpl", { ... })
  }
}

data "argocd_application" "apps" {
  for_each = toset(var.apps)
  metadata {
    name      = "${each.value}-${var.clustername}"
    namespace = "argocd"
  }
  depends_on = [null_resource.argocd_sync]
}

### Debug Output

module.sync_services.null_resource.argocd_sync: Creation complete after 47s [id=3955970134406545021]
module.sync_services.data.argocd_application.apps["kube-prometheus-stack"]: Reading...
╷
│ Error: Plugin did not respond
│
│ The plugin encountered an error, and failed to respond to the
│ plugin6.(*GRPCProvider).ReadDataSource call. The plugin logs may contain
│ more details.
╵

### Steps to Reproduce

1. Run a `local-exec` provisioner (e.g. via `null_resource`) that talks to the
   ArgoCD API (sync + wait), immediately followed by a `data "argocd_application"`
   read that depends_on that resource, using `for_each` over multiple app names.
2. `tofu apply`

### Expected Behavior

The data source read succeeds after the null_resource completes.

### Actual Behavior

The provider crashes with "Plugin did not respond" while reading the
data.argocd_application data source, immediately after the null_resource
creation completes. Downgrading the provider constraint from `~>7.0`
(which resolved to 7.16.0) to a pinned `7.15.3` resolves the issue reliably.

This looks similar in nature to #720 and #614 (crash / non-response during
concurrent provider calls shortly after a resource operation), but occurs
on a newer provider version (7.16.0, released 2 days prior to this report)
where no matching issue currently exists.

Thanks in advance for looking into this!

pujitha24:
Motivation:
The data.argocd_application data source (and the application resource's
read path) crashes the provider process ("Plugin did not respond") when an
application's spec.syncPolicy.automated block has any of its Prune,
SelfHeal, or AllowEmpty fields unset on the ArgoCD server side. This is a
very common situation: these three fields are independently optional, and
a user (or ArgoCD itself) may set only one or two of them.

The regression was introduced when the vendored argo-cd v3 client was
bumped to add v3.4 support. SyncPolicyAutomated.{Prune,SelfHeal,AllowEmpty}
changed from plain bool to *bool upstream, and
newApplicationSyncPolicyAutomated started dereferencing them
unconditionally (*spa.Prune, *spa.SelfHeal, *spa.AllowEmpty), causing a nil
pointer dereference panic whenever one of them is nil.

Approach:
Use the nil-safe GetPrune()/GetSelfHeal()/GetAllowEmpty() accessor methods
that the upstream argo-cd v1alpha1.SyncPolicyAutomated type already
provides for exactly this purpose (they return false when the receiver or
the field is nil), instead of dereferencing the pointers directly. This
restores the pre-regression behavior, where these fields defaulted to
false when unset.

Validation:
Added a targeted unit test
(internal/provider/model_application_test.go) that constructs a
SyncPolicyAutomated with all three pointer fields left nil and calls
newApplicationSyncPolicyAutomated directly. Confirmed this test panics
with "invalid memory address or nil pointer dereference" at the
now-fixed line against the pre-fix code, and passes cleanly after the
fix. Ran:
  USE_TESTCONTAINERS=false go test ./internal/provider/... -run TestNewApplicationSyncPolicyAutomated -v
and the broader:
  USE_TESTCONTAINERS=false go test ./internal/...
both pass. Also ran `go build ./...`, `go vet ./...`, and
`golangci-lint run ./internal/provider/...` (0 new issues; a handful of
pre-existing goconst warnings in unrelated lines/files are unchanged by
this diff).

Not run: this repo's `make testacc` acceptance suite (spins up a live
ArgoCD server in a K3s testcontainer) and `make generate`, because Docker
and the terraform CLI are not available in this environment. The change
only affects a value-conversion helper and does not touch any schema
definitions, so `make generate` should produce no diff, but that was not
verified by actually running it.

Report: https://github.com/argoproj-labs/terraform-provider-argocd/issues/919

Fixes #919

### DIFF
--- a/internal/provider/model_application.go
+++ b/internal/provider/model_application.go
@@ -928,9 +928,9 @@ func newApplicationSyncPolicyAutomated(spa *v1alpha1.SyncPolicyAutomated) *appli
 	}
 
 	return &applicationSyncPolicyAutomated{
-		AllowEmpty: types.BoolValue(*spa.AllowEmpty),
-		Prune:      types.BoolValue(*spa.Prune),
-		SelfHeal:   types.BoolValue(*spa.SelfHeal),
+		AllowEmpty: types.BoolValue(spa.GetAllowEmpty()),
+		Prune:      types.BoolValue(spa.GetPrune()),
+		SelfHeal:   types.BoolValue(spa.GetSelfHeal()),
 	}
 }
 
--- a/internal/provider/model_application_test.go
+++ b/internal/provider/model_application_test.go
@@ -1,0 +1,49 @@
+package provider
+
+import (
+	"testing"
+
+	"github.com/argoproj/argo-cd/v3/pkg/apis/application/v1alpha1"
+	"github.com/stretchr/testify/assert"
+)
+
+// A running application's SyncPolicy.Automated fields (Prune, SelfHeal,
+// AllowEmpty) are *bool and are frequently nil when not explicitly set by
+// the user, e.g. right after an out-of-band sync triggered outside Terraform.
+// newApplicationSyncPolicyAutomated must not dereference them directly.
+func TestNewApplicationSyncPolicyAutomated_NilFields(t *testing.T) {
+	t.Parallel()
+
+	spa := &v1alpha1.SyncPolicyAutomated{}
+
+	assert.NotPanics(t, func() {
+		result := newApplicationSyncPolicyAutomated(spa)
+
+		assert.False(t, result.AllowEmpty.ValueBool())
+		assert.False(t, result.Prune.ValueBool())
+		assert.False(t, result.SelfHeal.ValueBool())
+	})
+}
+
+func TestNewApplicationSyncPolicyAutomated_SetFields(t *testing.T) {
+	t.Parallel()
+
+	allowEmpty, prune, selfHeal := true, true, true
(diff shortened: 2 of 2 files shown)


onematchfox (approved):

the-technat (approved):
Thanks for fixing this! That was clearly my mistake...

PR #927, opened 2026-08-13
fix: return a diagnostic instead of panicking on malformed application import IDs

### LINKED ISSUE #529: Provider crashes when using import block. Index out of range.

```
Terraform version: v1.9.4
ArgoCD provider version: v7.0.3-7.2.0
ArgoCD version: v2.13.2
```

### Affected Resource(s)
- argocd_application when doing an import 

```
locals {
  argo_app_id = "cluster-bootstrap-${var.workload_name}-${var.env}"
}

import {
  to = argocd_application.bootstrap_cluster
  id = local.argo_app_id
}
```

```
resource "argocd_cluster" "eks" {
  server  = module.eks.cluster_endpoint
  name    = module.eks.cluster_name
  project = var.env

  config {
    aws_auth_config {
      cluster_name = module.eks.cluster_name
      role_arn     = var.argo_role_arn
    }
    tls_client_config {
      ca_data = base64decode(module.eks.cluster_certificate_authority_data)
    }
  }

  lifecycle {
    ignore_changes = [
      metadata
    ]
  }
}

resource "argocd_application" "bootstrap_cluster" {
  metadata {
    name      = "cluster-bootstrap-${var.workload_name}-${var.env}"
    namespace = "argocd"
  }

  spec {
    project = "${var.env}"

    destination {
      name      = "eks-${var.workload_name}-${var.env}"
      namespace = "${var.workload_name}"
    }

    source {
      repo_url        = "https://gitlab.com/breeze-airways/devops/argocd.git"
      path            = "helm/cluster-bootstrap"
      target_revision = "HEAD"

      helm {
        values = yamlencode({
          #destination_server      = ""
          destination_server_name = "eks-${var.workload_name}-${var.env}"
          project                = var.env
          cluster_endpoint       = module.eks.cluster_endpoint
          cluster_name          = module.eks.cluster_name
          env                   = var.env
          region               = var.region
          vpc_id               = data.aws_vpc.eks.id

          karpenter = {
            enabled      = true
            version      = "1.0.3"
            sa_role_arn = module.karpenter.iam_role_arn
            interruption_queue = "Karpenter-eks-${var.env}"
          }

          nops = {
            enabled    = true
            cluster_id = module.eks.cluster_id
          }

          alb_controller = {
            enabled      = true
            sa_role_arn = module.alb_controller_irsa_role.iam_role_arn
            version     = "1.7.2"
          }

          external_secrets = {
            enabled      = true
            sa_role_arn = var.deploy_external_secrets ? module.eso_irsa_role[0].iam_role_arn : null
          }

          hpa = {
            enabled = true
          }
        })
      }
    }
    sync_policy {
      automated {
        prune = true
        self_heal = true
      }
      sync_options = [
        "CreateNamespace=true"
      ]
    }
  }
}
```

### Debug Output
https://gist.github.com/dkstevens2000/25124344dcdb83a27aec950ecb2ee02a

### Panic Output
https://gist.github.com/dkstevens2000/6d5a1b4d91cb49da8deeb79a18e83df9

### Steps to Reproduce
Run terraform apply on the above configuration

### Expected Behavior
Should create an argo application called cluster-bootstrap 

### Actual Behavior
argocd provider crashes if the import is used.

### Community Note

pujitha24:
Motivation:
The argocd_application and argocd_application_set resources use
schema.ImportStatePassthroughContext, which sets the Terraform resource ID
to whatever string is supplied via `terraform import` or an `import` block,
without validation. resourceArgoCDApplicationRead/Update/Delete and
resourceArgoCDApplicationSetRead/Delete all did
`ids := strings.Split(d.Id(), ":"); name := ids[0]; namespace := ids[1]`,
which panics with "runtime error: index out of range [1] with length 1"
if the supplied ID has no colon. This happens in practice when an `import`
block is given a bare resource name (e.g. `id = local.app_name`) instead of
the documented `<name>:<namespace>` format, crashing the whole provider
process instead of returning a normal Terraform error.

Approach:
Added parseNameNamespaceID in argocd/utils.go, which splits the ID with
strings.SplitN(id, ":", 2) and returns a diag.Diagnostics error (matching
the existing argoCDAPIError/errorToDiagnostics conventions) when the ID
does not contain exactly two non-empty parts, instead of allowing an
out-of-range index panic. All five call sites that previously indexed
strings.Split(d.Id(), ":") directly (Read/Update/Delete for
argocd_application, Read/Delete for argocd_application_set) now use this
helper and return its diagnostics on error. User-visible behavior for
valid IDs (the documented `<name>:<namespace>` format) is unchanged; the
only behavior change is that a malformed ID now surfaces as a normal
Terraform error diagnostic instead of crashing the provider.

Validation:
- go build ./... passed.
- go vet ./argocd/... passed.
- golangci-lint run ./argocd/... : the only finding touching a changed
  file is a pre-existing goconst warning on an unrelated, untouched line.
- USE_TESTCONTAINERS=false go test ./argocd/... -run 'TestParseNameNamespaceID|TestValidatePolicy' -v
  passed, including a case reproducing the exact issue scenario (a bare
  name with no namespace segment).
- A standalone repro of the pre-fix code path
  (strings.Split("cluster-bootstrap-foo", ":"); ids[1]) reproduced the
  exact panic message from the issue's gist:
  "runtime error: index out of range [1] with length 1".
- Not run: full acceptance tests (make testacc), which require Docker to
  spin up a K3s + ArgoCD environment unavailable in this sandbox. This
  fix is a deterministic string-parsing defect fully provable via the
  unit test and standalone repro above, so a live-cluster run is not
  needed to establish correctness.
- Note: argoproj-labs/terraform-provider-argocd's main branch currently
  has some failing "Acceptance Tests" CI jobs unrelated to this change.

Report: https://github.com/argoproj-labs/terraform-provider-argocd/issues/529

Fixes #529

### DIFF
--- a/argocd/utils.go
+++ b/argocd/utils.go
@@ -99,10 +99,12 @@ func persistToState(key string, data interface{}, d *schema.ResourceData) error 
 // argocd_application_set use this ID format, which is set as-is on import
 // (e.g. via a `terraform import` block), so an ID missing the namespace
 // segment must be reported as a diagnostic rather than causing an
-// out-of-range panic when indexed.
+// out-of-range panic when indexed. The namespace segment may be empty
+// (e.g. 'myapp:') since ArgoCD installations without applicationset/app
+// multi-namespace support leave it unset.
 func parseNameNamespaceID(resource, id string) (name string, namespace string, diags diag.Diagnostics) {
 	ids := strings.Split(id, ":")
-	if len(ids) != 2 || ids[0] == "" || ids[1] == "" {
+	if len(ids) != 2 || ids[0] == "" {
 		return "", "", diag.Diagnostics{
 			{
 				Severity: diag.Error,
--- a/argocd/utils.go
+++ b/argocd/utils.go
@@ -99,12 +99,14 @@ func persistToState(key string, data interface{}, d *schema.ResourceData) error 
 // argocd_application_set use this ID format, which is set as-is on import
 // (e.g. via a `terraform import` block), so an ID missing the namespace
 // segment must be reported as a diagnostic rather than causing an
-// out-of-range panic when indexed. The namespace segment may be empty
-// (e.g. 'myapp:') since ArgoCD installations without applicationset/app
-// multi-namespace support leave it unset.
+// out-of-range panic when indexed. The namespace segment is optional and
+// defaults to empty (e.g. 'myapp' and 'myapp:' are equivalent) since an
+// empty namespace is treated the same as an unset one by the underlying
+// ArgoCD API.
 func parseNameNamespaceID(resource, id string) (name string, namespace string, diags diag.Diagnostics) {
 	ids := strings.Split(id, ":")
-	if len(ids) != 2 || ids[0] == "" {
+
+	invalid := func() (string, string, diag.Diagnostics) {
 		return "", "", diag.Diagnostics{
 			{
 				Severity: diag.Error,
@@ -114,7 +116,20 @@ func parseNameNamespaceID(resource, id string) (name string, namespace string, d
 		}
 	}
 
-	return ids[0], ids[1], nil
+	switch len(ids) {
+	case 1:
+		name = ids[0]
+	case 2:
+		name, namespace = ids[0], ids[1]
+	default:
+		return invalid()
+	}
+
+	if name == "" {
+		return invalid()
+	}
+
+	return name, namespace, nil
 }
 
 func argoCDAPIError(action, resource, id string, err error) diag.Diagnostics {
--- a/argocd/utils.go
+++ b/argocd/utils.go
@@ -101,7 +101,7 @@ func persistToState(key string, data interface{}, d *schema.ResourceData) error 
 // segment must be reported as a diagnostic rather than causing an
 // out-of-range panic when indexed.
 func parseNameNamespaceID(resource, id string) (name string, namespace string, diags diag.Diagnostics) {
-	ids := strings.SplitN(id, ":", 2)
+	ids := strings.Split(id, ":")
 	if len(ids) != 2 || ids[0] == "" || ids[1] == "" {
 		return "", "", diag.Diagnostics{
 			{
--- a/argocd/utils.go
+++ b/argocd/utils.go
@@ -3,6 +3,7 @@ package argocd
 import (
 	"fmt"
 	"strconv"
+	"strings"
 
 	"github.com/argoproj-labs/terraform-provider-argocd/internal/features"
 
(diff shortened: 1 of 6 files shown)


argocd/utils.go
```diff
@@ -180,6 +180,27 @@ func persistToState(key string, data interface{}, d *schema.ResourceData) error
...
 }
 
+// parseNameNamespaceID splits a resource ID of the form '<name>:<namespace>'
+// into its two parts. Resources such as argocd_application and
+// argocd_application_set use this ID format, which is set as-is on import
+// (e.g. via a `terraform import` block), so an ID missing the namespace
+// segment must be reported as a diagnostic rather than causing an
+// out-of-range panic when indexed.
+func parseNameNamespaceID(resource, id string) (name string, namespace string, diags diag.Diagnostics) {
+	ids := strings.SplitN(id, ":", 2)
```
onematchfox:
``` suggestion
	ids := strings.Split(id, ":")
```

Kubernetes namespaces cannot contain colons - neither can ArgoCD Applications so this function should return an error if `len(ids) != 2` rather than using `strings.SplitN(..., 2)` which will ensure `len(ids)` always equals 2. Test will need to updated accordingly as well.

onematchfox (changes requested):

onematchfox (dismissed):

argocd/utils_test.go
```diff
@@ -89,3 +89,69 @@ func TestValidatePolicy(t *testing.T) {
...
+			expectDiags: true,
+		},
+		{
+			name:        "Missing name",
+			id:          ":argocd",
+			expectDiags: true,
+		},
+		{
+			name:        "Empty namespace - as set for resources without multi-namespace support",
+			id:          "myapp:",
```
onematchfox, line 48: Glad the acceptance tests are doing their job and protecting against my poor memory. 😄 

I'm wondering if we need to change approach here. Functionally there is no difference between `myapp` and `myapp:`. So, rather than allowing one and not the other perhaps we should fix/protect against the nil pointer issue by defaulting namespace in both cases? Please also double-check the underlying library's handling of `""` vs `nil` for namespace?

the-technat: I checked the updated PR and now it seems like this case is catched, at least it LGTM. We might want to expand the import docs a bit though as the empty ns case is not mentioned at all, neither do we link to the Argo CD docs/behavior of what happens by submitting an app without a namespace. WDYT @onematchfox

onematchfox: Yeah, a doc update would probably be a good idea

onematchfox (approved):

the-technat (approved):
Thanks for fixing this @pujitha24, I'm planning on releasing the changes together with other stuff by the end of the week.

